PR #5, opened 2020-06-11
init Bazel WORKSPACE and create a dummy cc_binary

### LINKED ISSUE #2: Set up this repository as a Bazel WORKSPACE
@tengpengli Perhaps you can start with this first, and write a simple dummy cc_binary() rule in a BUILD file, just to get a basic development workflow going.

tengpengli:
Fixes #2 

Init Bazel WORKSPACE and create a hello world cc_binary


hello-world.cc
```diff
@@ -0,0 +1,8 @@
+#include <ctime>
```
stefanbucur, line 1: Add top-level comment documenting that this is an example file that will be replaced with the actual fuzz target rule functionality.

BUILD
```diff
@@ -0,0 +1,6 @@
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+
+cc_binary(
+    name = "hello-world",
+    srcs = ["hello-world.cc"],
```
stefanbucur, line 5: Should we use underscores as separators in file names? @asraa is there a specific Envoy convention that we could perhaps follow here?

BUILD
```diff
@@ -0,0 +1,6 @@
+load("@rules_cc//cc:defs.bzl", "cc_binary")
```
stefanbucur, line 1: Add a one-line description at the top. Perhaps:

```
Simple Bazel build file that will be replaced with the actual fuzz target rule functionality.
```

BUILD
```diff
@@ -0,0 +1,6 @@
...
+
+cc_binary(
+    name = "hello-world",
+    srcs = ["hello-world.cc"],
+)
```
stefanbucur, line 6: Should we have a `cc_test` too, that could be checked automatically by Travis as a PR check?

BUILD
asraa: We prefer underscores, but abide by Google style guide which says bother are ok so long as consistent: https://google.github.io/styleguide/cppguide.html#File_Names

BUILD
asraa: +1

BUILD
tengpengli: Vote for underscores. Personally I prefer underscores. The file names here cane from the Bazel document, I thought it's Bazel's tradition and thus copy & paste.

stefanbucur: Then maybe stick with `_`? I know there are standard directory names for Google projects that follow this convention (e.g., `third_party`).

.github/bazel_test.yml
```diff
@@ -0,0 +1,11 @@
+name: Bazel Test
```
asraa, line 1: Does this file need to be in .github/workflows/bazel_test.yml to trigger?

tengpengli: Oops, forgot to move to workflow/

.github/workflows/bazel_test.yml
```diff
@@ -0,0 +1,11 @@
+name: Bazel Test
+
+on: [push, pull_request]
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dummy test
```
stefanbucur, line 9: Maybe call this "Run all tests"?

.github/workflows/bazel_test.yml
```diff
@@ -0,0 +1,11 @@
...
+        run: |
+          bazel test //...
```
stefanbucur, line 11: The test seems to fail because it cannot find the workspace. Do you know which `cwd` this is running in?

tengpengli: As far as I can find online, the working-directory should be able to explicitly define the directory in which the command runs, but so far I failed with setting it as "." and "./"

asraa: Does the WORKSPACE need a line like
workspace(name = "bazel-rules-fuzzing")?

Once I hit this issue, and i used print commands in the github workflow for debugging :P

tengpengli: It seems adding that line to the WORKSPACE file doesn't work :(

tengpengli: > Run echo `pwd`
/home/runner/work/bazel-rules-fuzzing/bazel-rules-fuzzing

I print the working directory path. It seems the Actions creates a sub-directory with the same name to run the command.

asraa: > ERROR: The 'test' command is only supported from within a workspace (below a directory having a WORKSPACE file).


I wonder if this is happening because it's a top level test in the same directory as the WORKSPACE

asraa: Also maybe it's worth printing whats in the directory. I looked at my example worfklow, and I had an action to checkout the repo. 
```
    steps:
    - name: Checkout repository
      uses: actions/checkout@v2
```

tengpengli: I think the checkout does the magic. Now the bazel test can be successfully triggered.

Do you think I should revert recent commits to push a more clear update? I mean, cutting out those messy and debug commits.

.github/workflows/bazel_test.yml
```diff
@@ -0,0 +1,19 @@
+name: Bazel Test
+
+on: [push, pull_request]
```
asraa, line 3: Final question -- what does pull_request do in addition to push? I see two checks, and I'm guessing they might be redundant?

tengpengli: Good point! I guess no pull_request can skip the push step?

stefanbucur: Would that mean the Workflow would run after the PR is approved and before it's pushed? I guess we want that to run at PR creation time too.

stefanbucur: From https://help.github.com/en/actions/reference/events-that-trigger-workflows, it seems that the "push" event happens every time a commit is pushed to a repository branch. So if someone forks this repo and sends a PR from its clone, there won't be a local branch and the trigger might not run. Should we use `pull_request` instead?

tengpengli: I added `pull_request` back, is it necessary to delete `push`?

stefanbucur: It looks like the `push` can be deleted, unless you'd like to run the check for outstanding work happening in a branch before a PR is created.

tengpengli: I think the `push` is needed when people is developing on the feature branch/non-master branch, which means the `pull request` may not be created thus the tests won't be run if no `push` here.

stefanbucur: So either way (keep both or just pull_request) is fine with me. Leaving the choice up to you 😄

.github/workflows/bazel_test.yml
stefanbucur: Flattening the commits upon PR merging sounds good to me. The Github UI should handle that.

PR #27, opened 2020-07-08
suppress "uninitialized" warning for msan_fuzz_test

### LINKED ISSUE #25: Spurious -Wuninitialized warning during CI builds in the MSAN fuzz target
CI builds show a `warning: variable 'a' is uninitialized when used here [-Wuninitialized]`. Since this is intentional, we should silence the warning by adding a `-Wno-uninitialized` flag to the `copts` attribute of the fuzz target rule.

tengpengli:
Fixes #25

### DIFF
--- a/examples/BUILD
+++ b/examples/BUILD
@@ -41,6 +41,7 @@ cc_fuzz_test(
 cc_fuzz_test(
     name = "msan_fuzz_test",
     srcs = ["msan_fuzz_test.cc"],
+    copts = ["-Wno-uninitialized"],
 )
 
 cc_fuzz_test(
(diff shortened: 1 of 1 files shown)


examples/BUILD
```diff
@@ -41,6 +41,7 @@ cc_fuzz_test(
 cc_fuzz_test(
     name = "msan_fuzz_test",
     srcs = ["msan_fuzz_test.cc"],
+    copts = ["-Wno-uninitialized"],
```
stefanbucur, line 44: Perhaps add a comment explaining that the test is designed to trigger an uninitialized memory issue?

PR #31, opened 2020-07-09
Initial version of the Python launcher script for invoking fuzz target executables

### LINKED ISSUE #17: Implement a tool for executing a fuzzing binary
The goal of this extra level of indirection is to provide a uniform command line interface behind a number of arbitrary fuzzing engines. The tool can be implemented in Python, with flags specified using ABSL's flag library.

We can start with basic functionality such as performing a plain fuzzing run with a given timeout.

tengpengli:
Fixes #17 

This script aims at providing an uniform command line interface behind a number of arbitrary fuzzing engines. For now the script perform a plain fuzzing run with a given timeout.  

The command to use this script looks like
```
bazel run --config=asan-libfuzzer //examples:empty_fuzz_test --run_under='//fuzzing/tools:launcher --timeout=5'
```

### DIFF
--- a/fuzzing/tools/launcher.py
+++ b/fuzzing/tools/launcher.py
@@ -5,14 +5,12 @@ This is the launcher script to provide a uniform command line interface behind a
 
 from absl import app
 from absl import flags
-from subprocess import call, TimeoutExpired
+from subprocess import call, TimeoutExpired, STDOUT
 import os
 
 FLAGS = flags.FLAGS
 
 flags.DEFINE_integer('timeout', 20, 'test timeout.', lower_bound=0)
-flags.DEFINE_enum('engine', 'libfuzzer', ['libfuzzer'],
-                  'the fuzzing engine used to do fuzzing test.')
 
 
 def main(argv):
@@ -20,10 +18,9 @@ def main(argv):
         raise app.UsageError("Too few arguments")
 
     ret_code = 0
-    cwd = os.getcwd()
     try:
         # Is this the right way to execute the runnable?
-        ret_code = call(cwd + "/../../../" + argv[1], timeout=FLAGS.timeout)
+        ret_code = call(argv[1], timeout=FLAGS.timeout)
     except TimeoutExpired as e:
         print("Error: Timeout")
     except Exception as e:
--- a/fuzzing/tools/launcher.py
+++ b/fuzzing/tools/launcher.py
@@ -10,16 +10,20 @@ import os
 
 FLAGS = flags.FLAGS
 
-flags.DEFINE_integer('timeout', 20, 'test timeout.', lower_bound=0)
+flags.DEFINE_integer(
+    'timeout_secs',
+    20,
+    'The maximum duration, in seconds, of the fuzzer run launched.',
+    lower_bound=0)
 
 
 def main(argv):
     if len(argv) != 2:
         raise app.UsageError(
-            "This script receives 2 arguments. It should look like:" +
-            "\n\tpython " + __file__ + " EXECUTABLE --timout=TIMEOUT")
+            "This script receives 1 argument. It should look like:" +
+            "\n\tpython " + __file__ + " EXECUTABLE")
 
-    os.execl(argv[1], argv[1], "-timeout=" + str(FLAGS.timeout))
+    os.execl(argv[1], argv[1], "-timeout=" + str(FLAGS.timeout_secs))
 
 
 if __name__ == '__main__':
--- a/fuzzing/tools/launcher.py
+++ b/fuzzing/tools/launcher.py
@@ -1,0 +1,36 @@
+# Lint as: python3
+"""
+This is the launcher script to provide a uniform command line interface behind a number of arbitrary fuzzing engines.
+"""
+
+from absl import app
+from absl import flags
+from subprocess import call, TimeoutExpired
+import os
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_integer('timeout', 20, 'test timeout.', lower_bound=0)
+flags.DEFINE_enum('engine', 'libfuzzer', ['libfuzzer'],
+                  'the fuzzing engine used to do fuzzing test.')
+
+
+def main(argv):
+    if len(argv) < 2:
+        raise app.UsageError("Too few arguments")
+
+    ret_code = 0
+    cwd = os.getcwd()
+    try:
+        # Is this the right way to execute the runnable?
+        ret_code = call(cwd + "/../../../" + argv[1], timeout=FLAGS.timeout)
+    except TimeoutExpired as e:
+        print("Error: Timeout")
+    except Exception as e:
+        print("Error: " + str(e))
+    if ret_code:
+        print("Error: " + str(ret_code))
+
+
+if __name__ == '__main__':
+    app.run(main)
--- a/fuzzing/tools/launcher.py
+++ b/fuzzing/tools/launcher.py
@@ -1,11 +1,11 @@
 # Lint as: python3
 """
-This is the launcher script to provide a uniform command line interface behind a number of arbitrary fuzzing engines.
+This is the launcher script to provide a uniform command line interface 
+behind a number of arbitrary fuzzing engines.
 """
 
 from absl import app
 from absl import flags
-from subprocess import call, TimeoutExpired, STDOUT
 import os
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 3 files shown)


fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,36 @@
...
+    ret_code = 0
+    cwd = os.getcwd()
+    try:
+        # Is this the right way to execute the runnable?
+        ret_code = call(cwd + "/../../../" + argv[1], timeout=FLAGS.timeout)
+    except TimeoutExpired as e:
+        print("Error: Timeout")
+    except Exception as e:
+        print("Error: " + str(e))
+    if ret_code:
```
asraa, line 31: Will this end up swallowing the crash and stack trace from the fuzzing call?

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,36 @@
...
+
+from absl import app
+from absl import flags
+from subprocess import call, TimeoutExpired
+import os
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_integer('timeout', 20, 'test timeout.', lower_bound=0)
+flags.DEFINE_enum('engine', 'libfuzzer', ['libfuzzer'],
```
asraa, line 14: for simplicity since this isn't used i think let's remove it? in the end a fuzzing engine will be specified with an individual launcher script, correct?

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,36 @@
...
+
+def main(argv):
+    if len(argv) < 2:
+        raise app.UsageError("Too few arguments")
+
+    ret_code = 0
+    cwd = os.getcwd()
+    try:
+        # Is this the right way to execute the runnable?
+        ret_code = call(cwd + "/../../../" + argv[1], timeout=FLAGS.timeout)
```
asraa, line 26: Hmmmmmm i was under the impression that you wouldn't need cwd/../../../ and that just invoking the binary will work ok.

fuzzing/tools/launcher.py
tengpengli: Yes, it will. One way to enable the output is passing an stream to stderr and stdout in call(). I think we can add another argument for launcher.py to be the log file name?

tengpengli: I double checked, if the executable is not terminated by the script, the message should be okay to be printed out to STDOUT.

fuzzing/tools/launcher.py
tengpengli: Make sense to me!

fuzzing/tools/launcher.py
tengpengli: Can I mark this conversation as resolved? Or should I add a stream argument into the launcher now?

asraa: What did you mean by "I double checked, if the executable is not terminated by the script, the message should be okay to be printed out to STDOUT."?

I think the launcher should definitely print out the error and stack trace, but also given that this is a slightly different setup that the end-goal, it's possible that it will change to do so when the script isn't run with CLI --run_under

tengpengli: I mean, when the timeout argument is big enough to let the fuzzing binary stop by itself, the output will contains error and stack trace, otherwise the output only contains "Error: Timeout".

stefanbucur:
Sorry for the latency, crazy week :) Starting to look at the PR now.

To begin with, I would suggest starting to include more descriptive information in the PR title and description, given that this is a public repository of a longer-term project and all this text will be persisted here.

As a rule of thumb, you should treat commit messages and PR descriptions as technical writing: Use proper grammar, full sentences, capitalization, punctuation, spell checking, etc.

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,33 @@
+# Lint as: python3
+"""
+This is the launcher script to provide a uniform command line interface behind a number of arbitrary fuzzing engines.
```
stefanbucur, line 3: Maybe wrap all source files at 80 character lines? (You should also check your editor - there is a good chance it can do this for you automatically.)

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,33 @@
...
+import os
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_integer('timeout', 20, 'test timeout.', lower_bound=0)
+
+
+def main(argv):
+    if len(argv) < 2:
+        raise app.UsageError("Too few arguments")
```
stefanbucur, line 18: I'm not sure the user will know how to use the script if they only get a `Too few arguments` error. Perhaps turn this into something more actionable? Maybe `Missing fuzzer binary path argument.`

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,33 @@
...
+from subprocess import call, TimeoutExpired, STDOUT
+import os
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_integer('timeout', 20, 'test timeout.', lower_bound=0)
+
+
+def main(argv):
+    if len(argv) < 2:
```
stefanbucur, line 17: What happens if `len(argv) > 2`? Should we trigger an error there, too?

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,33 @@
...
+        raise app.UsageError("Too few arguments")
+
+    ret_code = 0
+    try:
+        # Is this the right way to execute the runnable?
```
stefanbucur, line 22: > Is this the right way to execute the runnable?

Really good question. I guess `call` will also work here, but then you will need to manage the subprocess itself: decide how to interpret exit statuses, propagate output, etc.

Since this is a launcher tool, I think it is safe to simply cede control of the execution to the fuzzer binary. You can use the `os.exec*` system calls, which essentially replaces the Python executable with the command passed as its argument.

@asraa WDYT?

fuzzing/tools/launcher.py
stefanbucur: Please see my other comment for a suggestion of a simpler alternative - using `os.exec` here.

fuzzing/tools/launcher.py
tengpengli: I used `yapf --style google ` to format the code, does the command look good to you?

stefanbucur: I haven't used it before, so can't really tell :/ But I know the style guide suggests wrapping at 80 chars: https://google.github.io/styleguide/pyguide.html#32-line-length

tengpengli: Got it, thank you!

I am wondering if you are using any tool to format you python code? If so, can you share that with me?

stefanbucur: I've only been using internal tools, but I don't think they can be made available for open source projects. I'd still stand by my original suggestion - look into your editor's documentation to see how to do it.

fuzzing/tools/launcher.py
tengpengli: Is there anyway to timeout the process when using `os.exec*`, or any parameter to control that when running the binary?

stefanbucur: Yes, libFuzzer provides the `-timeout` argument: https://llvm.org/docs/LibFuzzer.html#options

tengpengli: Awesome! I was struggling with this for quite a while!

fuzzing/tools/launcher.py
tengpengli: Changed to
```
This script receives 2 arguments. It should look like:
        python fuzzing/tools/launcher.py EXECUTABLE --timout=TIMEOUT```

File: fuzzing/tools/launcher.py [hunk 8]
```diff
@@ -0,0 +1,26 @@
...
+behind a number of arbitrary fuzzing engines.
+"""
+
+from absl import app
+from absl import flags
+import os
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_integer('timeout', 20, 'test timeout.', lower_bound=0)
```
stefanbucur, line 13: Flags should be documented more thoroughly. Here someone seeing `test timeout` will likely not understand what this is about. I'll give a suggestion here, and expect you'll apply the same principles for future flags:

```
The maximum duration, in seconds, of the fuzzer run launched.
```

stefanbucur, line 13: Call this flag `timeout_secs`, to clarify the unit of measure on the command line?

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def main(argv):
+    if len(argv) != 2:
+        raise app.UsageError(
+            "This script receives 2 arguments. It should look like:" +
```
stefanbucur, line 19: I believe this is incorrect: The script receives one argument only, the other one being argv[0], which is the executable path.

fuzzing/tools/launcher.py
```diff
@@ -0,0 +1,26 @@
...
+            "\n\tpython " + __file__ + " EXECUTABLE --timout=TIMEOUT")
```
stefanbucur, line 20: `--timeout` is an optional flag, it shouldn't appear in the error message. Moreover, this doesn't scale as more flags will be added.

PR #34, opened 2020-07-15
Initial version of cc_fuzz_run, which creates a script <name>_run

### LINKED ISSUE #18: Augment the cc_fuzz_test() macro with a new <name>_run rule
The _run rule should invoke the fuzzer runner tool (#17) when run using "bazel run ...".

tengpengli:
**Commit Message:**
Initial version of cc_fuzz_run, which creates a script \<name\>_run
The <name>_run calls the launcher script to run the \<name\> executable
**Additional Description:**
The only difference between this PR and #31 is the cc_deps.bzl.
The command inside the generated script is absolutely not what we are looking for, but without timeout, the bazel run never exists. How can I make the script end? And how to pass the timeout_secs to the launcher script?
**Testing:**
CI testing and local testing.
**Docs Changes:** N/A
**Fixes #18**

### DIFF
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-""" This file contains basic functions for cc fuzz test. """
+"""This file contains basic functions for cc fuzz test."""
 
 load("@rules_cc//cc:defs.bzl", "cc_test")
 load("//fuzzing:common.bzl", "fuzzing_launcher")
@@ -29,7 +29,7 @@ def cc_fuzz_test(
     """This macro provide two targets:
     <name>: the executable file built by cc_test.
     <name>_run: an executable to launch the fuzz test.
-    """
+"""
 
     cc_test(
         name = name,
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -13,53 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-""" This file contains basic functions for fuzz test. """
+""" This file contains basic functions for cc fuzz test. """
 
 load("@rules_cc//cc:defs.bzl", "cc_test")
-
-def _cc_fuzz_run_impl(ctx):
-    # Generate a script to launcher the fuzzing test
-    script = ctx.actions.declare_file("%s" % ctx.label.name)
-
-    # This command is absolutely not what we are looking for,
-    # but without timeout here, the bazel run never exists.
-    # What is wrong here?
-    script_template = """#!/bin/bash
-timeout 10 {launcher} {target}
-"""
-
-    script_content = script_template.format(
-        launcher =
-            ctx.attr._launcher[DefaultInfo].files_to_run.executable.short_path,
-        target =
-            ctx.attr.dep[DefaultInfo].files_to_run.executable.short_path,
-    )
-    ctx.actions.write(script, script_content, is_executable = True)
-
-    # Merge the two dependencies
-    runfiles = ctx.attr._launcher[DefaultInfo].default_runfiles
-    runfiles = runfiles.merge(ctx.attr.dep[DefaultInfo].default_runfiles)
-
-    return [DefaultInfo(executable = script, runfiles = runfiles)]
-
-# What should the argument be? How to pass timeout_secs?
-cc_fuzz_run = rule(
-    implementation = _cc_fuzz_run_impl,
-    attrs = {
-        # The launcher script to start fuzzing test
-        "_launcher": attr.label(
-            default = Label("//fuzzing/tools:launcher"),
-            executable = True,
-            cfg = "host",
-        ),
-        # The _fuzz_test executable to run
-        "dep": attr.label(
-            executable = True,
-            cfg = "host",
-        ),
-    },
-    executable = True,
-)
+load("//fuzzing:common.bzl", "fuzzing_launcher")
 
 def cc_fuzz_test(
         name,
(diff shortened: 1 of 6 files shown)


fuzzing/cc_deps.bzl
```diff
@@ -17,6 +17,50 @@
...
+    )
+    ctx.actions.write(script, script_content, is_executable = True)
+
+    # Merge the two dependencies
+    runfiles = ctx.attr._launcher[DefaultInfo].default_runfiles
+    runfiles = runfiles.merge(ctx.attr.dep[DefaultInfo].default_runfiles)
+
+    return [DefaultInfo(executable = script, runfiles = runfiles)]
+
+# What should the argument be? How to pass timeout_secs?
```
stefanbucur, line 45: Re. the `--timeout_secs` question, wouldn't something like this work?

```
$ bazel run --config=asan-libfuzzer //path/to:fuzzer_run -- --timeout_secs=60
```

fuzzing/cc_deps.bzl
```diff
@@ -17,6 +17,50 @@
...
+        # The launcher script to start fuzzing test
+        "_launcher": attr.label(
+            default = Label("//fuzzing/tools:launcher"),
+            executable = True,
+            cfg = "host",
+        ),
+        # The _fuzz_test executable to run
+        "dep": attr.label(
+            executable = True,
+            cfg = "host",
```
stefanbucur, line 58: Should this be "target" instead? I guess we want the `asan-libfuzzer` configuration and all other build settings to apply on this one.

fuzzing/cc_deps.bzl
```diff
@@ -17,6 +17,50 @@
...
+    runfiles = ctx.attr._launcher[DefaultInfo].default_runfiles
+    runfiles = runfiles.merge(ctx.attr.dep[DefaultInfo].default_runfiles)
+
+    return [DefaultInfo(executable = script, runfiles = runfiles)]
+
+# What should the argument be? How to pass timeout_secs?
+cc_fuzz_run = rule(
+    implementation = _cc_fuzz_run_impl,
+    attrs = {
+        # The launcher script to start fuzzing test
```
stefanbucur, line 49: .bzl source code comments will not be visible to developers using this rule. Consider turning them into `doc` fields for each attribute defined. This is documented in more depth here: https://github.com/bazelbuild/stardoc/blob/master/docs/writing_stardoc.md

fuzzing/cc_deps.bzl
```diff
@@ -17,6 +17,50 @@
...
+    ctx.actions.write(script, script_content, is_executable = True)
+
+    # Merge the two dependencies
+    runfiles = ctx.attr._launcher[DefaultInfo].default_runfiles
+    runfiles = runfiles.merge(ctx.attr.dep[DefaultInfo].default_runfiles)
+
+    return [DefaultInfo(executable = script, runfiles = runfiles)]
+
+# What should the argument be? How to pass timeout_secs?
+cc_fuzz_run = rule(
```
stefanbucur, line 46: I imagine this launcher is not specific to C++ fuzz targets, so perhaps call this rule something more generic? I'd suggest `fuzzing_launcher` (and similarly, `_fuzzing_launcher_impl` for the rule implementation).

fuzzing/cc_deps.bzl
asraa: Agreed, passing args through the CLI probably works in this case., and mimics normal fuzzing invocations. 

I think you could also add an explicit arg for a timeout to the rule, or maybe other ways, but the above usage would be good (and I think should work).

fuzzing/cc_deps.bzl
tengpengli: With that name, do you think it will be more proper to place this rule into common.bzl?

stefanbucur: Good point, yes, placing it in common.bzl makes a lot of sense.

tengpengli:
I think I found the reason why my <name>_run script never ends.

In my local test, I was using empty_fuzz_test_run to test the script. The `--timeout_secs` argument at the very end is passed as the libfuzzer argument `-timeout`, which set the time limit for **one round of fuzz test**. In the meantime, the empty_fuzz_test  function simply returns, which ends really quickly, thus the timeout would never be triggered. In other words, the script never ends.

stefanbucur:
You're absolutely right! It was the wrong flag to use, didn't even notice that :) It seems like https://llvm.org/docs/LibFuzzer.html has also a `-max_total_time` flag: can you try that and see if it works?

tengpengli:
> You're absolutely right! It was the wrong flag to use, didn't even notice that :) It seems like https://llvm.org/docs/LibFuzzer.html has also a `-max_total_time` flag: can you try that and see if it works?

It works for empty_fuzz_run, but the interesting thing is that the oom_fuzz_test and hang_fuzz_test don't get interrupted. 

I only tried those three, maybe it's because the `-max_total_time` doesn't set a timer for the child thread created by the fuzzing test, but only a timer on the parent thread? 

Anyway, I use both `-timeout` and `-max_total_time`, now it runs as expected.

tengpengli:
PTAL!

For the timeout problem in the launcher script, I fixed it in #35

fuzzing/cc_deps.bzl
```diff
@@ -13,9 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-""" This file contains basic functions for fuzz test. """
+""" This file contains basic functions for cc fuzz test. """
```
stefanbucur, line 16: Nit: Remove the extra space between the quotes and the text?

PR #37, opened 2020-07-16
Test an example _run targets in CI

### LINKED ISSUE #36: Add _run targets to CI
## Expected Behavior
Test an example _run targets with a short timeout (eg 5 seconds) in CI

## Actual Behavior
Add something like `bazel run //examples:empty_fuzz_test_run --config=asan-libfuzzer -- --timeout_secs=5` to CI

tengpengli:
**Commit Message:**
Test an example _run targets in CI
**Additional Description:**
Test an example empty_fuzz_test_run targets with a short timeout (eg 5 seconds) in CI
**Testing:**
CI testing and local testing.
**Docs Changes:** N/A
**Fixes #36**

### DIFF
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -31,3 +31,4 @@ jobs:
         # The "<fuzzer/FDP...>" is only supported since clang-10 while the clang version in github runner is 9
         run: |
           bazel build --verbose_failures --build_tag_filters=fuzz_test,-FDP --config=asan-libfuzzer //examples/...
+          bazel run //examples:empty_fuzz_test_run --config=asan-libfuzzer -- --timeout_secs=5
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -25,10 +25,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Install deps 
         run: |
-          pip --version
           sudo apt-get install python3-setuptools
-          sudo apt-get install python-setuptools
-          pip install wheel absl-py
           pip3 install wheel absl-py
       - name: Run all tests
         working-directory: ./
@@ -37,7 +34,6 @@ jobs:
         # The -FDP is for filtering out the files which contains "#include <fuzzer/FuzzedDataProvider.h>",
         # The "<fuzzer/FDP...>" is only supported since clang-10 while the clang version in github runner is 9
         run: |
-          python --version
           alias python='python3'
           bazel build --verbose_failures --build_tag_filters=fuzz_test,-FDP --config=asan-libfuzzer //examples/...
           bazel run //examples:empty_fuzz_test_run --config=asan-libfuzzer -- --timeout_secs=5
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -23,6 +23,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - name: Install deps 
+        run: |
+          pip install absl-py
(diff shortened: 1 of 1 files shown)


tengpengli:
Have problem installing the absl-py. Since this is not in the critical path, I will pend this issue for a little while and come back to fix this later

.github/workflows/bazel_test.yml
```diff
@@ -31,3 +35,4 @@ jobs:
         # The "<fuzzer/FDP...>" is only supported since clang-10 while the clang version in github runner is 9
         run: |
           bazel build --verbose_failures --build_tag_filters=fuzz_test,-FDP --config=asan-libfuzzer //examples/...
+          bazel run //examples:empty_fuzz_test_run --config=asan-libfuzzer -- --timeout_secs=5
```
tengpengli, line 38: What version of python is used when running _run?

.github/workflows/bazel_test.yml
```diff
@@ -23,6 +23,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - name: Install deps 
+        run: |
+          pip install wheel absl-py
+          pip3 install wheel absl-py
```
tengpengli, line 29: Check the version of pip and see if it's the same version as python

.github/workflows/bazel_test.yml
tengpengli: The `python` command in github runner is python2. I have to declare `alias python="python3"` to run python3

.github/workflows/bazel_test.yml
tengpengli: Default `pip` is `pip2`.

tengpengli:
The codes are ready to serve. PTAL!

.github/workflows/bazel_test.yml
```diff
@@ -23,11 +23,17 @@ jobs:
...
       - name: Run all tests
         working-directory: ./
         # Build //examples/...
         # Add "bazel test --config=clang //... --test_tag_filters=test"in the future if any test rules are added
         # The -FDP is for filtering out the files which contains "#include <fuzzer/FuzzedDataProvider.h>",
         # The "<fuzzer/FDP...>" is only supported since clang-10 while the clang version in github runner is 9
         run: |
+          alias python='python3'
           bazel build --verbose_failures --build_tag_filters=fuzz_test,-FDP --config=asan-libfuzzer //examples/...
+          bazel run //examples:empty_fuzz_test_run --config=asan-libfuzzer -- --timeout_secs=5
```
stefanbucur, line 39: It's great to test this command; I was wondering what we can do here to make the test fail if something other than a build goes wrong. Some failure scenarios in mind:

* The timeout is not obeyed, causing the test to run indefinitely. How will we catch this error? Can we make the command fail automatically (say if we leave it to run for 5 seconds, forcefully terminate it after 30 seconds) so it doesn't take too long to run? @asraa did Envoy have to come up with a solution to this problem?
* The command runs, but doesn't produce the expected output. For example, if we run a crashing fuzz target, running that command should expose the crash.

For the second problem, I imagine writing separate shell scripts and invoking them from here seems more scalable.

In any case, we can still push what we have now, and open a separate issue for improving it.

asraa: We could add a timeout to the test script as a whole (like 5-10 min), which won't give fine-grained control over each test runtime, but would catch something like a fuzzer is running indefinitely 
(https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes)

For the second, I agree, having a framework to say "run this, expect with grep some error line comes up" would be cool

PR #38, opened 2020-07-16
Rule fuzzing_corpus for collecting corpus files

### LINKED ISSUE #12: Define a fuzzing_corpus() Starlark rule
The rule will be used as part of the main `cc_fuzz_test` macro. Example usage:

```bzl
fuzzing_corpus(
    name = "my_fuzzer_corpus",
    srcs = [
        "file1.txt",
        "path/to/file2.txt",
    ],
)
```

tengpengli:
**Commit Message:**
Rule fuzzing_corpus for collecting corpus files
**Additional Description:**
Example usage:

fuzzing_corpus(
    name = "my_fuzzer_corpus",
    srcs = [
        "file1.txt",
        "path/to/file2.txt",
    ],
)
**Testing:**
Local testing and CI testing
**Docs Changes:** N/A
**Fixes #12**

### DIFF
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -21,6 +21,7 @@ load("//fuzzing:common.bzl", "fuzzing_corpus", "fuzzing_launcher")
 
 def cc_fuzz_test(
         name,
+        corpus = [],
         **kwargs):
     """Macro for c++ fuzzing test
 
@@ -45,7 +46,7 @@ def cc_fuzz_test(
         testonly = True,
     )
 
-    if "corpus" in kwargs:
+    if corpus:
         fuzzing_corpus(
             name = name + "_corpus",
             srcs = kwargs["corpus"],
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -16,6 +16,7 @@
 """This file contains basic functions for cc fuzz test."""
 
 load("@rules_cc//cc:defs.bzl", "cc_test")
+load("@rules_pkg//:pkg.bzl", "pkg_zip")
 load("//fuzzing:common.bzl", "fuzzing_corpus", "fuzzing_launcher")
 
 def cc_fuzz_test(
@@ -55,5 +56,9 @@ def cc_fuzz_test(
     if corpus:
         fuzzing_corpus(
             name = name + "_corpus",
+            srcs = corpus,
+        )
+        pkg_zip(
+            name = name + "_corpus_zip",
             srcs = corpus,
         )
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -17,7 +17,6 @@
 
 load("@rules_cc//cc:defs.bzl", "cc_test")
 load("//fuzzing:common.bzl", "fuzzing_corpus", "fuzzing_launcher")
-load("@rules_pkg//:pkg.bzl", "pkg_zip")
 
 def cc_fuzz_test(
         name,
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -21,7 +21,7 @@ load("//fuzzing:common.bzl", "fuzzing_corpus", "fuzzing_launcher")
 
 def cc_fuzz_test(
         name,
-        corpus = [],
+        corpus = None,
         **kwargs):
     """Macro for c++ fuzzing test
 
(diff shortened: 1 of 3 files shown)


tengpengli:
I am using linux command `cp` to collect the corpus file. In my local test, the `<name>_corpus` is only available to collect the files under `examples/` and the its sub-directories. I am not sure that is what we want. Even it is, I am still curious about how to collect the files outside the directory of the target.

From my experiments, absolute path like "/usr/home/path/to/file.txt", which starts with '/', is not allowed. A relative path like "../parent/dir/file.txt" is also not acceptable. I am wondering what is the right way to pass the path of a file in the other directories?

tengpengli:
Apologize for keeping pushing new little updates.

PTAL now!

tengpengli:
In #39 we use kwargs to forward attributes to cc_test in cc_fuzz_test. Since `cc_test` doesn't accept `corpus` but we need `corpus` here, we need to delete `corpus` entry in `**kwargs` after calling `launcher_corpus`

tengpengli:
Added <name>_corpus_zip rule to create the zip file. PTAL!

fuzzing/common.bzl
```diff
@@ -56,3 +56,32 @@ Rule for creating a script to run the fuzzing test
     },
     executable = True,
 )
+
+def _fuzzing_corpus_impl(ctx):
+    dir = ctx.actions.declare_directory(ctx.attr.name)
```
stefanbucur, line 61: We should use more explicit variable names to help the reader go through this code more easily. I'll leave suggestions below, feel free to apply the same principle for new code:

`dir` -> `corpus_dir`
`f` -> `input_file`

fuzzing/cc_deps.bzl
```diff
@@ -43,3 +44,13 @@ def cc_fuzz_test(
         # this attribute must be set.
         testonly = True,
     )
+
+    if "corpus" in kwargs:
+        fuzzing_corpus(
```
asraa, line 49: I'd love to see this passed in to the `fuzzing_launcher` in a later iteration.

So that if I invoke the run target, I start the libfuzzer process with this corpus as a seed corpus by passing it as an arg to the script.

fuzzing/cc_deps.bzl
```diff
@@ -43,3 +44,13 @@ def cc_fuzz_test(
         # this attribute must be set.
         testonly = True,
     )
+
+    if "corpus" in kwargs:
```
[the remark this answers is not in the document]
tengpengli: I agree on adding the corpus as a arguments, but I am wondering if you can explain the difference between `corpus = None` and
 ```
...
corpus = [],
**kwargs):
...
if corpus:
...
```
 which can skip `corpus = corpus or []`?

File: fuzzing/cc_deps.bzl (hunk 2 above)
Comment (tengpengli) [reply fuzzing/cc_deps.bzl:49]: I have created #44, and plan to support this feature in the next week if that is okay to you.

**Comment by tengpengli** [comment]
> Tengpeng, could you add a test (or example corpus for an example fuzz test) for this? I think after building a `cc_fuzz_target`, you should be able to expect a corpus in bazel's output directory.

Of course! Do you mean adding an target into `examples/` with `corpus` argument not being empty, and then making a CI test for building this target?

File: fuzzing/cc_deps.bzl [hunk 4]
```diff
@@ -43,3 +45,13 @@ def cc_fuzz_test(
         # this attribute must be set.
         testonly = True,
     )
+
+    if corpus:
+        fuzzing_corpus(
+            name = name + "_corpus",
+            srcs = kwargs["corpus"],
```
stefanbucur, line 52: Stale use of `kwargs` here and below?

fuzzing/cc_deps.bzl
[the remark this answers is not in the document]
stefanbucur: Ah, good question! The difference is a subtle one. It has to do with how macro arguments are interpreted by the Starlark language. From the [Conventions section](https://docs.bazel.build/versions/master/skylark/macros.html#conventions) in the docs:

> In most cases, optional parameters should have a default value of None. [...] Additionally, a parameter that is explicitly set to its default value looks different than one that is never set (or set to None) when accessed through the query language or build-system internals.

asraa:
> Of course! Do you mean adding an target into examples/ with corpus argument not being empty, and then making a CI test for building that target?

Yep, either with a new target or with an existing target. (It could also be part of #44, but I think an example test or usage of this feature would be needed. For eg, specifying separate individual files, specifying all files in a subdirectory, or both)

fuzzing/cc_deps.bzl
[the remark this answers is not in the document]
tengpengli: Thank you for the explanation and the doc link! 

I have one more question. After setting the default value as `None`, I am wondering why we need `corpus = corpus or []`? In python, we will never enter `if None` block, which I think is what we want. Is there anything I missed?

PR #39, opened 2020-07-16
Use **kwargs to forward attributes in cc_fuzz_test

### LINKED ISSUE #26: The cc_fuzz_test macro should forward all attributes it doesn't handle.
Right now we explicitly forward attributes, e.g., `native.cc_test(..., deps = deps, ...)`. The macro should receive instead a `**kwargs` argument that captures all attributes that we want to pass through, and forward that to the `cc_test` rule instantiation.

tengpengli:
**Commit Message:**
Forward all attributes that `cc_fuzz_test` can't handle to `cc_test` by
using keyword arguments `**kwargs`.
**Additional Description:**
Since `cc_test` doesn't accept `corpus` but we may pass `corpus` in `cc_fuzz_test`, we need to delete `corpus` entry in `**kwargs` after calling `launcher_corpus` in #38 
**Testing:**
Local testing and CI testing
**Docs Changes:** N/A
**Fixes #26**

### DIFF
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -23,7 +23,7 @@ def cc_fuzz_test(
         **kwargs):
     """Macro for c++ fuzzing test
 
-    This macro provide two targets:
+    This macro provides two targets:
     <name>: the executable file built by cc_test.
     <name>_run: an executable to launch the fuzz test.
 """
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -21,8 +21,7 @@ load("//fuzzing:common.bzl", "fuzzing_launcher")
 def cc_fuzz_test(
         name,
         srcs,
-        **kwargs,
-        ):
+        **kwargs):
     """Macro for c++ fuzzing test
 
     This macro provide two targets:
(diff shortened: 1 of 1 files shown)


fuzzing/cc_deps.bzl
```diff
@@ -21,24 +21,21 @@ load("//fuzzing:common.bzl", "fuzzing_launcher")
...
+        **kwargs):
+    """Macro for c++ fuzzing test
+
+    This macro provide two targets:
     <name>: the executable file built by cc_test.
     <name>_run: an executable to launch the fuzz test.
 """
 
+    # Add fuzz_test tag
+    kwargs["tags"] = ["fuzz_test"] + (kwargs["tags"] if "tags" in kwargs else [])
```
stefanbucur, line 33: I think this can be simplified to `kwargs.setdefault("tags", []).append("fuzz_test")`.

fuzzing/cc_deps.bzl
```diff
@@ -21,24 +21,21 @@ load("//fuzzing:common.bzl", "fuzzing_launcher")
 def cc_fuzz_test(
         name,
         srcs,
```
stefanbucur, line 23: Should we allow `src` to be unspecified? Presumably, sometimes the fuzz target source may already be available as a library, so we just need to link that. In that case, we can also drop the `srcs` argument here.

PR #45, opened 2020-07-23
Add corpus as an argument to fuzzing_launcher

### LINKED ISSUE #44: Add fuzzing_corpus as an argument to fuzzing_launcher
## Expected Behavior
The `<name>_run` target should be able to be invoked with a seed corpus, which is passed as the `corpus` argument in the `cc_fuzz_test`. 

We have already supported copying the corpus files  to `<name>_corpus` directory by the `fuzzing_corpus` rule. So this feature can be implemented by passing the `_corpus` directory as an argument into the command in the `_run` script.

An example as [discussed](https://github.com/googleinterns/bazel-rules-fuzzing/pull/38#issuecomment-661901573) to show the usage of this feature is also needed.

## Actual Behavior
The `fuzzing_launcher` is not able to accepts the seed corpus as an argument now.

tengpengli:
**Commit Message:**
Add corpus as an argument to fuzzing_launcher
**Additional Description:**
Make fuzzing_corpus a runfile of fuzzing_launcher.
Add corpus_dir to launcher.py, which is the directory containing corpus files.
Add corpus to empty_fuzz_test in BUILD as an example of the new feature.
**Testing:**
Local testing and CI testing
**Docs Changes:** N/A
**Fixes #44**

	modified:   examples/BUILD
	new file:   examples/corpus_1.txt
	new file:   examples/corpus_dir/corpus_2.txt
	modified:   fuzzing/tools/launcher.py


examples/BUILD
```diff
@@ -22,9 +22,14 @@ cc_fuzz_test(
     srcs = ["buffer_overflow_fuzz_test.cc"],
 )
 
+# This target shows how to create a fuzz test target with corpus files
```
stefanbucur, line 25: Retrofitting an existing target for corpus support creates confusion (and the fact that we need a comment to clarify it is another sign of it IMO).

How about creating a new cc_fuzz_test() target called "empty_fuzz_test_with_corpus" that re-uses the "empty_fuzz_test.cc" source file, but adds corpus support?

fuzzing/cc_deps.bzl
```diff
@@ -55,3 +47,19 @@ def cc_fuzz_test(
...
+            testonly = True,
+        )
+    else:
+        fuzzing_launcher(
+            name = name + "_run",
+            target = name,
+            # Since the script depends on the _fuzz_test above, which is a cc_test,
+            # this attribute must be set.
+            testonly = True,
+        )
```
stefanbucur, line 65: We should avoid code duplication as much as possible. The fact that we even copy-pasted the exact code comments is another strong sign that this should be factored out.

How about instantiating `fuzzing_launcher` once and only setting the `corpus` attribute when needed, like this:

```
fuzzing_launcher(
  name = name + "_run",
  ...,
  corpus = name + "_corpus" if corpus else None,
)
```

Note the use of "None" to specify a `corpus` attribute that is not set - this is the way Starlark signals it.

fuzzing/common.bzl
```diff
@@ -20,18 +20,22 @@ def _fuzzing_launcher_impl(ctx):
     script = ctx.actions.declare_file("%s" % ctx.label.name)
 
     script_template = """#!/bin/sh
-exec {launcher_path} {target_binary_path} "$@"
+exec {launcher_path} {target_binary_path} --corpus_dir={corpus_dir} "$@"
 """
 
+    print(ctx.file.corpus.short_path)
```
stefanbucur, line 26: Remove? Otherwise all users will see these as warning/debug messages.

fuzzing/common.bzl
```diff
@@ -74,7 +82,7 @@ def _fuzzing_corpus_impl(ctx):
         command = "cp $@",
     )
 
-    return [DefaultInfo(files = depset([corpus_dir]))]
+    return [DefaultInfo(runfiles = ctx.runfiles(files = [corpus_dir]), files = depset([corpus_dir]))]
```
stefanbucur, line 85: This is getting a bit complex on one line, perhaps break it such that `runfiles = ...` and `files = ...` are on separate lines?

fuzzing/tools/launcher.py
```diff
@@ -11,23 +26,31 @@
...
 
 flags.DEFINE_integer(
-    'timeout_secs',
+    "timeout_secs",
     20,
-    'The maximum duration, in seconds, of the fuzzer run launched.',
+    "The maximum duration, in seconds, of the fuzzer run launched.",
     lower_bound=0)
 
+flags.DEFINE_string("corpus_dir", "", "The directory containing corpus.")
```
stefanbucur, line 34: We can make the documentation of the flag more precise to suggest it is optional: `If non-empty, a directory that will be used as a seed corpus for the fuzzer run.`

fuzzing/tools/launcher.py
```diff
@@ -11,23 +26,31 @@
...
+    if FLAGS.corpus_dir:
+        os.execv(argv[1], [
+            argv[1], FLAGS.corpus_dir, "-max_total_time=" +
+            str(FLAGS.timeout_secs), "-timeout=" + str(FLAGS.timeout_secs)
+        ])
+    else:
+        os.execv(argv[1], [
+            argv[1], "-max_total_time=" + str(FLAGS.timeout_secs),
+            "-timeout=" + str(FLAGS.timeout_secs)
+        ])
```
stefanbucur, line 52: Same comment about duplication as above. Consider instead constructing the argument list separately, and using conditionals to append arguments to the list. Something like this:

```py
command_line = [argv[1]]
command_line.append("-max_total_time=" + str(FLAGS.timeout_secs))
command_line.append("-timeout=" + str(FLAGS.timeout_secs))
if FLAGS.corpus_dir:
  command_line.append(...)
os.execv(argv[1], command_line)
```

examples/BUILD
```diff
@@ -27,6 +27,15 @@ cc_fuzz_test(
     srcs = ["empty_fuzz_test.cc"],
 )
 
+cc_fuzz_test(
+    name = "empty_fuzz_test_with_corpus",
+    srcs = ["empty_fuzz_test.cc"],
+    corpus = [
+        "corpus_1.txt",
+        "corpus_dir/corpus_2.txt",
```
asraa, line 35: nit: small ask, can you demonstrate specifying a whole directory of files like `corpus_dir/**` with 1+ files? We often have a whole corpus, and want to specify that

fuzzing/common.bzl
```diff
@@ -20,25 +20,28 @@ def _fuzzing_launcher_impl(ctx):
...
 
     script_template = """#!/bin/sh
-exec {launcher_path} {target_binary_path} "$@"
+exec {launcher_path} {target_binary_path} --corpus_dir={corpus_dir} "$@"
 """
 
     script_content = script_template.format(
         launcher_path = ctx.executable._launcher.short_path,
         target_binary_path = ctx.executable.target.short_path,
+        corpus_dir = ctx.file.corpus.short_path if ctx.attr.corpus else "",
```
asraa, line 29: because this is "" if there's no corpus, won't the launcher script always have this flag?

tengpengli: Yes, the flag will always be there, and in the launcher script, the default value of `corpus_dir` is also `""`(a empty string), which will be ignored later when passing the argument to the target binary to execute. Do you think I should add a judgement here to decide whether to pass the `corpus_dir`?

asraa: I think handling it with a default empty string is nice, so that the launcher script can always except a corpus string (even if empty). I think my question is, won't the launcher script work fine even with an empty string for the corpus? 
In this case it should just run with no corpus, libfuzzer will still work without any corpus

tengpengli: In my experiment, when appending `""` to `execv` argument list, it causes an error `No such file or directory: ; exiting`. Seems even passing an empty string, it's regarded as an directory path, which makes the `if` statement in the launcher script  necessary.

asraa: Awesome, thank you for clarifying and testing that!

PR #46, opened 2020-07-23
Fuzzing dictionary validation python script

### LINKED ISSUE #15: Implement a fuzzing dictionary validation tool
We should use Python for the tool. We should assume AFL & libFuzzer's dictionary format, as documented here:
 * The general dictionary format: https://github.com/rc0r/afl-fuzz/tree/master/dictionaries.
 * The libFuzzer [implementation of dictionary parsing](https://github.com/llvm/llvm-project/blob/6c67ee0f58323fe39db84635ef8cea629c2214b4/compiler-rt/lib/fuzzer/FuzzerUtil.cpp#L127), which only extracts the words between quotes in each dictionary line. We can implement the parser as a regular expression.

We should also add a few tests for the tool, which parse a few sample dictionary files.

tengpengli:
**Commit Message:**
Fuzzing dictionary validation python script
**Additional Description:**
Example usage:
```python dict_validator.py DICTIONARY_FILE```

The code models [libfuzzer dictionary parser](https://github.com/llvm/llvm-project/blob/6c67ee0f58323fe39db84635ef8cea629c2214b4/compiler-rt/lib/fuzzer/FuzzerUtil.cpp#L81)

The validator will regard `"abc""` as an valid entry, which I am not sure whether it should be or not.
**Testing:**
Local testing and CI testing
**Docs Changes:** N/A
**Fixes #15**

	new file:   dict_validator.py

### CODE REVIEW

File: fuzzing/tools/dict_validator.py [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+            return False
+
+        # Handle '\\'
+        if chr == '\\':
+            if left + 1 <= right and (line[left + 1] == '\\' or
+                                      line[left + 1] == '"'):
+                left += 1
+                continue
+
+            # Handle '\xAB'
```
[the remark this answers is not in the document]
tengpengli: I am wondering if we have any existed tools validating the dictionary entry? 

The codes here are modeling  [libfuzzer dictionary parser](https://github.com/llvm/llvm-project/blob/6c67ee0f58323fe39db84635ef8cea629c2214b4/compiler-rt/lib/fuzzer/FuzzerUtil.cpp#L81). It allows `"ab""` to pass the check while filters out `"\xA"`, those look strange to me.

fuzzing/tools/dict_validator.py
```diff
@@ -0,0 +1,89 @@
...
+
+from absl import app
+from absl import flags
+from string import hexdigits
+
+FLAGS = flags.FLAGS
+
+
+# Validate a single entry in the dictionary
+def validate_entry(line):
```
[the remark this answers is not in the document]
tengpengli: Create #47 for the regular expression replacement.

fuzzing/tools/dict_validator_test.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+Tests for dict_validator.py
+"""
+
+from string import hexdigits
+from dict_validator import validate_line
+import sys
+
+if __name__ == '__main__':
+    with open("dict_data/valid.dict", 'r') as dic:
```
tengpengli, line 26: The file is already added to `data` attribute in `BUILD`. However, when running `bazel test`, the test failed with `FileNotFoundError`. Under the `runfiles` directory, there did exist those files. 

I also tried `os.path.join` and `$TEST_SRCDIR` to get the absolute path like [this envoy python test](https://github.com/envoyproxy/envoy/blob/defea7ecbf6f3ebffe8b9d41afa7f62322756f05/api/tools/generate_listeners_test.py#L8), but didn't work. I am wondering how to enable the `py_test` to find the data files? 

The same error for the `invalid.dict` below.

asraa: Couple questions, what was the `$TEST_SRCDIR` environment variable? What is the value of `os.path.dirname(__file__)`? I think getting runfiles is a little tricky with Bazel and python....

asraa: In Envoy for C++ we use a bazel runfiles library, there seems to be one for Python as well
https://github.com/bazelbuild/bazel/blob/4e4c77ae96955138006af15664074123d6a4f51d/tools/python/runfiles/runfiles.py#L16

tengpengli: Thank you Asra!

I tried to `print` out the `$TEST_SRCDIR` variable and `dirname(__file__)`, but can't see the output in the terminal, neither in the `test.log`. Then I tried the `runfiles.Rlocation` in the link, it helped the bazel test find the test data!

PTAL!

fuzzing/tools/dict_validator.py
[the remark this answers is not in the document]
asraa: Yep, libfuzzer's own dictionary parser is probably the gold standard. You can test out some of these cases by running any fuzzer with `-dict`, and it will tell you if there are any dictionary errors, and how many successful dictionary entries there were. I tried the case above, and found
```-----------------------------------------------------------------------------
ParseDictionaryFile: error in line 2
		"\xA"
```

It is definitely worth adding these testcases to the test dictionaries, and running a fuzzer with the test dictionaries to make sure the valid ones are parsed, and each invalid one isn't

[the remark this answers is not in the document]
tengpengli: Yes, I have created `valid.dict` and `invalid.dict` for the valid and invalid dictionary entries respectively under `fuzzing/tools/dict_data`.

For the `"ab""` with which I was confused, it turns out to be a valid entry.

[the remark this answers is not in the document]
asraa: Ah yeah, sorry -- I meant adding "\xA" to the `valid.dict` and also just verifying that all of valid.dict gets accepted to libfuzzer by running the fuzzer with it!

fuzzing/tools/dict_data/valid.dict
```diff
@@ -0,0 +1,27 @@
...
+# Lines starting with '#' and empty lines are ignored.
+
+# Adds "blah" (w/o quotes) to the dictionary.
+kw1="blah"
+# Use \\ for backslash and \" for quotes.
+kw2="\"ac\\dc\""
+# Use \xAB for hex values
+kw3="\xF7\xF8""
+# the name of the keyword followed by '=' may be omitted:
+"foo\x0Abar"
```
asraa, line 27: add "ab""

tengpengli:
I created a new target `dict_validator_unit_test`, which uses python `unittest` module to assert the result. 

AFAICS, the new `_unit_test` works similar to the previous `dict_validator_test`, do you think it 's okay to only keep the `_unit_test` target?

fuzzing/tools/dict_data/invalid.dict
```diff
@@ -0,0 +1,3 @@
+Invalid dictionary entries
```
stefanbucur, line 1: This could probably be turned into separate test cases in the unit test.

fuzzing/tools/BUILD
```diff
@@ -15,10 +15,43 @@
 
 # This package contains tool scripts
 
-load("@rules_python//python:defs.bzl", "py_binary")
+load("@rules_python//python:defs.bzl", "py_binary", "py_library", "py_test")
+
+py_library(
+    name = "dict_validator_lib",
+    srcs = ["dict_validator.py"],
```
stefanbucur, line 22: Let's have a separate .py file with the library, and a separate tool file that provides the binary for the Bazel rule.

fuzzing/tools/BUILD
```diff
@@ -15,10 +15,43 @@
...
+    visibility = ["//visibility:public"],
```
stefanbucur, line 23: Does this really need to be public?

fuzzing/tools/BUILD
```diff
@@ -15,10 +15,43 @@
...
+    ],
+)
+
+py_test(
+    name = "dict_validator_unit_test",
+    srcs = ["dict_validator_unit_test.py"],
+    deps = [
+        ":dict_validator_lib",
+    ],
+)
```
stefanbucur, line 57: It doesn't look like we need both of these - let's consolidate our testing in the API unit test.

fuzzing/tools/dict_validator.py
```diff
@@ -0,0 +1,99 @@
...
+    if left >= len(line) - 1:
+        return False
+
+    return validate_entry(line[left:])
+
+
+def main(argv):
+    with open(FLAGS.output_file, 'w') as output:
+        for dic_path in FLAGS.dict_list:
+            print(dic_path)
```
stefanbucur, line 85: What is the role of this print() statement?

fuzzing/tools/dict_validator.py
```diff
@@ -0,0 +1,99 @@
...
+            with open(dic_path, 'r') as dic:
+                for line in dic.readlines():
+                    if not validate_line(line):
+                        print("ERROR: invalid dictionary entry \'" +
```
stefanbucur, line 89: Errors should be printed to stderr by convention.

fuzzing/tools/dict_validator.py
```diff
@@ -0,0 +1,99 @@
...
+                              line.strip() + "\'")
+                        return -1
+                    output.write(line)
+
+    print("The dictionary is successfully validated.")
```
stefanbucur, line 94: All this output will be printed by Bazel during builds - it is a bit noisy and perhaps unnecessary? (Imagine if `clang` would output "Compilation successful" for each .c file compiled :).

fuzzing/tools/dict_validator_test.py
```diff
@@ -0,0 +1,41 @@
...
+                      "' can't pass the check")
+                sys.exit(-1)
+
+    with open(rf.Rlocation("__main__/fuzzing/tools/dict_data/invalid.dict"),
+              'r') as dic:
+        for line in dic.readlines():
+            if validate_line(line):
+                print("ERROR: invalid dictionary entry '" + line.strip() +
+                      "' can pass the check")
+                sys.exit(-1)
```
stefanbucur, line 41: Let's turn these into `unittest` test cases and merge into the unit test file.

fuzzing/tools/dict_validator_unit_test.py
```diff
@@ -0,0 +1,59 @@
...
+# Use \\xAB for hex values
+kw3="\\xF7\\xF8""
+# the name of the keyword followed by '=' may be omitted:
+"foo\\x0Abar"
+"ab""
+"""
+        for line in valid_lines.split('\n'):
+            # if not validate_line(line):
+            # print(line)
+            self.assertTrue(validate_line(line))
```
stefanbucur, line 48: Let's turn these into distinct test cases, with appropriate names. For example,

```
def test_plain_word(self):
  ...
def test_escaped_word(self):
  ...
def test_hex_escapes(self):
  ...
```

Why do this? Because if any of these cases fail, the unit testing framework will specifically point to the failing test case. Otherwise, each failure would be bundled under a single test case which would be harder to diagnose.

fuzzing/tools/dict_validator_unit_test.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_invalid_line(self):
```
stefanbucur, line 50: Same here - separate test cases.

fuzzing/tools/dict_validate_lib_test.py
```diff
@@ -0,0 +1,70 @@
...
+        lines = """kw3="\\xF7\\xF8""
+"foo\\x0Abar" """
+        for line in lines.split('\n'):
+            self.assertTrue(validate_line(line))
+        invalid_hex_str = '"\\A"'
+        self.assertFalse(validate_line(invalid_hex_str))
+
+    def test_comment(self):
+        line = """# valid dictionary entries"""
+        self.assertTrue(validate_line(line))
```
stefanbucur, line 50: Fold into single statement? I'm not sure I see the benefit of having a separate variable here.

tengpengli:
This PR involves 4 files, I am wondering whether you think it will be better to create another PR for the test in the future if we are in the similar situation?

stefanbucur:
I believe it is good practice to combine features and unit tests in the same PR. By not delaying writing them, we minimize the time code is submitted without tests.

fuzzing/tools/dict_validation.py
```diff
@@ -0,0 +1,88 @@
...
+        True if the argument is a valid fuzzing dictionary entry, 
+        otherwise False.
+    """
+
+    # Use set to contain hex digits to decrease the query time complexity
+    hex_set = set(hexdigits)
+    pos, end = 0, len(entry) - 1
+    while pos < end:
+        pos += 1
+        chr = entry[pos]  # Single character
```
stefanbucur, line 41: See observation above: Redundant comment?

fuzzing/tools/validate_dict.py
```diff
@@ -0,0 +1,51 @@
...
+from absl import flags
+from dict_validation import validate_line
+from sys import stderr
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_list("dict_list", [],
+                  "Each element in the list stands for a dictionary file")
+
+flags.DEFINE_string("output_file", "output.dict",
```
stefanbucur, line 31: Maybe start with an empty value for this flag? Otherwise users of this tool may be unaware that this file gets written and might overlap existing content.

fuzzing/tools/validate_dict.py
```diff
@@ -0,0 +1,51 @@
...
+    with open(FLAGS.output_file, 'w') as output:
+        for dic_path in FLAGS.dict_list:
+            with open(dic_path, 'r') as dic:
+                for line in dic.readlines():
+                    if not validate_line(line):
+                        print("ERROR: invalid dictionary entry \'" +
+                              line.strip() + "\'",
+                              file=stderr)
+                        return -1
+                    output.write(line)
```
stefanbucur, line 45: Maybe break this into one or more functions? There are currently 6 levels of indentation, which feels a bit too much.

stefanbucur:
BTW, I also don't see these tests actually running in our checks. Can you also address #50?

fuzzing/tools/validate_dict.py
```diff
@@ -0,0 +1,61 @@
...
+FLAGS = flags.FLAGS
+
+flags.DEFINE_list("dict_list", [],
+                  "Each element in the list stands for a dictionary file")
+
+flags.DEFINE_string("output_file", "",
+                    "The name of the output merged dictionary file")
+
+
+def validate_dic(dic_path, output_stream):
```
stefanbucur, line 35: I see "dictionary" being shortened to both "dict" and "dic" in this PR. Prefer being consistent - just stick to "dict" everywhere?

PR #51, opened 2020-08-03
Break up "Run all tests" in bazel_test.yml

### LINKED ISSUE #50: Break up the "Run all tests" step in bazel_test.yml
Currently, it is not possible to easily see what tests are being run in the "Run all tests" stage of the bazel_test.yml Github action. To fix this, we should break this step up in several steps:
1. One for building all fuzz test examples,
1. one for running all the other {cc,py}_tests, and
1. another section for the dedicated tests (for now, it's just the *_run invocation).

tengpengli:
**Commit Message:**
Break up "Run all tests" in bazel_test.yml

The original tests are divided into 3 parts:
1) Build fuzzing exmample targets under examplers/
2) Build and run all tests, i.e., targets with "test" tag
3) Run dedicated tests, for now, only _run targtet

**Additional Description:**
**Testing:**
Local testing and CI testing
**Docs Changes:** N/A
** Fixes #50**

	modified:   fuzzing/tools/BUILD


tengpengli:
This PR is making CI tests for `dict_validation` based on the branch `dic_valid`.

I am not sure if this is the right way to create a new feature based on a waiting-to-merge feature. Let me know if you think this way is improper or not efficient.

.github/workflows/bazel_test.yml
```diff
@@ -27,13 +27,15 @@ jobs:
         run: |
           sudo apt-get install python3-setuptools
           pip3 install wheel absl-py
-      - name: Run all tests
-        working-directory: ./
-        # Build //examples/...
-        # Add "bazel test --config=clang //... --test_tag_filters=test"in the future if any test rules are added
+          alias python='python3'
+      - name: Build examples
```
stefanbucur, line 31: Maybe be more specific: "Build fuzz test examples"

.github/workflows/bazel_test.yml
```diff
@@ -27,13 +27,15 @@ jobs:
...
         # The -FDP is for filtering out the files which contains "#include <fuzzer/FuzzedDataProvider.h>",
         # The "<fuzzer/FDP...>" is only supported since clang-10 while the clang version in github runner is 9
         run: |
-          alias python='python3'
           bazel build --verbose_failures --build_tag_filters=fuzz_test,-FDP --config=asan-libfuzzer //examples/...
+      - name: Run common tests
+        run: |
+          bazel test --config=clang //... --test_tag_filters=test --build_tag_filters=test
```
stefanbucur, line 38: Why are the `--test_tag_filters` and `--build_tag_filters` flags necessary here?

fuzzing/tools/BUILD
```diff
@@ -15,10 +15,38 @@
...
+    srcs = ["dict_validation.py"],
+)
+
+# Tests.
+########
+
+py_test(
+    name = "dict_validation_test",
+    srcs = ["dict_validation_test.py"],
+    tags = ["test"],
```
stefanbucur, line 50: I would like to understand why this is needed. I feel it is unsustainable to expect everyone to remember to add this tag to their unit tests.

.github/workflows/bazel_test.yml
tengpengli: The scenario in my mind is that in the future we may need the CI to run more tests like this. If we don't use the tag filters, instead we test the specific target here, then this file need to update every time a new test is added.

It is my assumption that the benefit getting from no need to update this CI script frequently is more than the potential risk that we may forget to add the tag. If my assumption is not correct or there is a way to achieve the goal without the tag filters, I absolutely support to delete the filters.

fuzzing/tools/BUILD
tengpengli: I am not sure which may have higher risk, forgetting to add this tag that leads to no CI test on the target or keeping changing the CI test script.

My original opinion is that keeping updating CI test script may cause more severe problem. That impression comes from Github requiring the authentication token to update the CI script, which is not needed in other place.

asraa: Maybe https://docs.bazel.build/versions/master/command-line-reference.html#flag--build_tests_only will help? I agree that manual tagging will be unsustainable

fuzzing/tools/validate_dict.py
```diff
@@ -32,9 +32,15 @@
                     "The name of the output merged dictionary file")
 
 
+<<<<<<< HEAD
```
asraa, line 35: please fix

tengpengli: Sorry, I fixed this locally, but forgot to save and upload.

fuzzing/tools/BUILD
asraa: Edit: another thing that can be done is make a placeholder `_test` fule for the "bazel test some_fuzz_test" that will be able to be filled in with a unit test implementation. 

(I think https://github.com/googleinterns/bazel-rules-fuzzing/blob/2cd72332724c1b1a4eca9edbc85a5f46cc60567a/fuzzing/cc_deps.bzl#L36 will have to become a `cc_binary`, so that it is not a `_test` rule, and a seperate `fuzzing_regression_test` rule needs to be made

tengpengli: I have a maybe stupid question, why we need `cc_binary` instead of `_test`? Is it just to avoid cases like this?

.github/workflows/bazel_test.yml
tengpengli: Used `--build_tests_only` instead.

PR #54, opened 2020-08-07
Explicitly use file.path in fuzzing_corpus

tengpengli:
**Commit Message:**
Explicitly use file.path in fuzzing_corpus

There is a CI error in Envoy mentioned the input_file may be a
directory, which can't be added by ctx.actions.args.add. As a workaround,
the path of the file is added instead.

**Additional Description:**
The shell command `cp $@` can add a `-r` parameter to avoid the file is a directory, though I am not sure that matches the user's desire and our plan.
**Testing:**
Local testing and CI testing.
**Docs Changes:** N/A

### DIFF
--- a/fuzzing/common.bzl
+++ b/fuzzing/common.bzl
@@ -69,7 +69,7 @@ def _fuzzing_corpus_impl(ctx):
     cp_args = ctx.actions.args()
 
     for input_file in ctx.files.srcs:
-        cp_args.add(input_file)
+        cp_args.add(input_file.path)
 
     # Add destination to the arguments
     cp_args.add(corpus_dir.path)


asraa:
I think this must be because of this special case of a corpus in Envoy https://github.com/envoyproxy/envoy/blob/2187f1070a76a124a4d7cebe72551804b1a5218b/test/common/router/BUILD#L224

It is constructed from a generated directory and a directory path. It goes to the "else" branch of https://github.com/envoyproxy/envoy/blob/be719cc0ebb3fb974ce45e32d6ba07060b874f1f/bazel/envoy_test.bzl#L91

tengpengli:
I tested locally (on my cloudtop), the filegroup can be passed to `fuzzing_corpus` like `[FILEGROUP]`, which can be done by the user. Does it mean this PR is not necessary? Or should we figure out a way to support both **filegroup** and **file path**?

stefanbucur:
AFAIK, the `filegroup` rule should still provide a set of files as its output, so it should be transparent to the users of those files whether they originated from a list of files or other transitive deps. 

Tengpeng, can you clarify the following part:

> There is a CI error in Envoy mentioned the input_file may be a
> directory, which can't be added by ctx.actions.args.add. As a workaround,
> the path of the file is added instead.

It's good to hear that this fixes the issue, but I'd like to understand why this fixes the issue.

From the [Bazel docs](https://docs.bazel.build/versions/master/skylark/lib/Args.html#modules.Args), "File objects are turned into their File.path values."

tengpengli:
In the [CI error message](https://dev.azure.com/cncf/envoy/_build/results?buildId=47517&view=logs&j=8d802004-fbbb-5f17-b73e-f23de0c1dec8&t=86512d3a-caa5-5e3f-9bce-4070faa3b830&l=119), it mentioned the problem can be solved by `Either use Args#add_all (if you want expansion) or args.add(directory.path) (if you do not)`, that's why I add `.path` in this PR. 

In addition, since the file may be a directory, I was also considering to add `-r` argument to the `cp` command, which was described in the PR description above.

Giving it a second thought, I think it will be better to claim the format of the input (i.e., a list of files) rather than change the command here. I am wondering what is your opinion? @asraa @stefanbucur

fuzzing/common.bzl
```diff
@@ -69,7 +69,7 @@ def _fuzzing_corpus_impl(ctx):
     cp_args = ctx.actions.args()
 
     for input_file in ctx.files.srcs:
-        cp_args.add(input_file)
+        cp_args.add(input_file.path)
```
stefanbucur, line 72: Could you also add a test that would fail before this change is introduced? This seems like a tricky case, and we probably want to have a regression test for this :)

asraa: Even better, a test with a `filegroup` as well would be awesome.

asraa:
> Giving it a second thought, I think it will be better to claim the format of the input (i.e., a list of files) rather than changing the command here. Keeping it this way, it can avoid accidentally moving unwanted files. I am wondering what is your opinion? @asraa @stefanbucur

Do you mean that the input should be declared as a list of files? 

> AFAIK, the filegroup rule would still provide a set of files as its output, so it should be transparent to the users of those files whether they originated from a list of files or other transitive deps.

I see. It is definitely possible to unwind the set of files that are created from the filegroup rule, but it's slightly opaque to the user of the genrule used to create the directory: 
https://github.com/envoyproxy/envoy/blob/374dca7905fc048be74169a7655d0462606555ad/test/common/router/BUILD#L192-L219
I would prefer being able to use both filegroups and lists of files. WDYT?

tengpengli:
> Do you mean that the input should be declared as a list of files?

Yes, I meant to let the user guarantee the input should be a list of either filegroups or files, which are supported now.

tengpengli:
> Does this mean that this PR alone would not fix the issue? What will now happen with paths that are directories? Will they be copied properly?

I am afraid this PR alone, i.e., without `-r` argument to `cp`, still can't fix the problem. According to the documentation you shared earlier, this PR is useless. 

In the `//examples:empty_fuzz_test_with_corpus`, I passed `"corpus_dir/**`, which will be expanded by `native.glob()` in `cc_fuzz_test`.  I am considering whether we should use `glob` directly in `empty_fuzz_test_with_corpus` rather than in `cc_fuzz_test`. The present codes make `cc_fuzz_test(..., corpus=glob(["corpus_dir/**",]) + [":filegroup"])` failed to run.
 
> Can you also comment on the Args#add_all option? Was this ruled out for a particular reason?

The option is not ruled out, just picked one of the two options.

> 
> I'm not sure I understand the "claim the format of the input" part. Can you elaborate?

I meant to let the user guarantee the input is a list of either filegroups or files.

tengpengli:
Tracking down the CI error message, I found the directory comes from the [":corpus_from_config_impl"](https://github.com/envoyproxy/envoy/blob/374dca7905fc048be74169a7655d0462606555ad/test/common/router/BUILD#L214), of which the target seems to return a directory and is passed to the fuzzing_launcher. Once it is changed to `glob([":corpus_from_config_impl"])`, the error is gone.

I am wondering if adding that `glob` will affect the result envoy wants? @asraa

asraa:
I think this is the idea. Globbing should be up to the user, and not involved in the rules fuzzing. For example, this test should succeed if you remove the globbing done in the `cc_fuzz_test` macro:

```
diff --git a/examples/BUILD b/examples/BUILD
index b44d7aa..25edf73 100644
--- a/examples/BUILD
+++ b/examples/BUILD
@@ -27,16 +27,20 @@ cc_fuzz_test(
     srcs = ["empty_fuzz_test.cc"],
 )
 
+filegroup(
+    name = "example_filegroup",
+    srcs = [
+        "corpus_1.txt",
+    ] + glob(["corpus_dir/**"]),
+)
+
 # This target shows how to create a fuzz test target with corpus files
 cc_fuzz_test(
     name = "empty_fuzz_test_with_corpus",
     srcs = ["empty_fuzz_test.cc"],
-    corpus = [
-        "corpus_1.txt",
-        # Single * only searches the files under that directory
-        # whereas double * will do the search recursively
-        "corpus_dir/**",
-    ],
+    corpus = [":example_filegroup"],
 )

```

tengpengli:
I think the case you gave should work with present codes. My local test looks exactly like yours:
```cc_fuzz_test(
    name = "empty_fuzz_test_with_corpus",
    srcs = ["empty_fuzz_test.cc"],
    corpus = [":corpus_filegroup"],
    # [
    #     "corpus_1.txt",
    #     # Single * only searches the files under that directory
    #     # whereas double * will do the search recursively
    #     "corpus_dir/**",
    # ],
)

filegroup(
    name = "corpus_filegroup",
    srcs = ["corpus_1.txt"] + glob(["corpus_dir/**"])
)
```

But for envoy, the error comes from  the [envoy_directory_genrule](https://github.com/envoyproxy/envoy/blob/49f6966c777ef1fa43f2d86a2ef590dd3928aa2e/bazel/envoy_build_system.bzl#L55) ":corpus_from_config_impl", which in my understanding will return a directory rather than some specific files. That's why I am asking whether `fuzzing_corpus` should support a argument like `[FILE_1, FILE_2,..., GROUPFILE_1,... DIRECTORY]`. If yes, is there anyway to tell the type of the element in the argument?

asraa:
> will return a directory rather than some specific files.

I see, so filegroup doesn't expand directories? https://github.com/envoyproxy/envoy/blob/49f6966c777ef1fa43f2d86a2ef590dd3928aa2e/test/common/router/BUILD#L224

tengpengli:
From my debug log, it seems not. The result of the `:corpus_from_config_impl` is `<generated file test/common/router/corpus_from_config_impl.outputs>`, which is still a directory. The full expanded list can be found [here](https://docs.google.com/document/d/1Pv6ME208R_sbCMQbCo4oSuYXibCGLL_6Cck9fkC_FFQ/edit?usp=sharing) if you are interested.

fuzzing/common.bzl
tengpengli: Created a new PR #57 for the test, PTAL!

tengpengli:
Since the problem discussed here is solved in #57 , can I close this PR?

stefanbucur:
Sure, closing it sounds good.

PR #56, opened 2020-08-08
Supports regression test in fuzzing_launcher

### LINKED ISSUE #55: Enable regression support in the launcher script
## Expected Behavior
Launcher needs to support running regression test without continuous fuzzing test. 

To achieve this, a string_flag `engine` will be added to decide the launcher's behavior. After the launcher is modified, a new rule `regression_launcher` will be needed to start the launcher in the regression mode.

## Actual Behavior
Only continuous fuzzing test mode is supported

tengpengli:
**Commit Message:**

Supports regression test in fuzzing_launcher

Adds string_flag engine to decide the fuzzing_launcher behavior.

**Additional Description:**
**Testing:**
Local testing and CI testing.
**Docs Changes:** N/A
**Fixes #55**

	modified:   .bazelrc
	modified:   WORKSPACE
	modified:   fuzzing/BUILD
	modified:   fuzzing/repositories.bzl
	modified:   fuzzing/tools/launcher.py


fuzzing/common.bzl
```diff
@@ -15,19 +15,22 @@
...
-"""
+exec {launcher_path} {target_binary_path} --corpus_dir={corpus_dir} "$@" """
 
     script_content = script_template.format(
         launcher_path = ctx.executable._launcher.short_path,
         target_binary_path = ctx.executable.target.short_path,
         corpus_dir = ctx.file.corpus.short_path if ctx.attr.corpus else "",
     )
+    if ctx.attr._engine[BuildSettingInfo].value == "default":
+        script_content += " --regression=True"
```
stefanbucur, line 33: I'm not sure this is right: The engine and whether it runs in regression mode or not should be orthogonal. We should also be able to run libFuzzer in regression mode, for example.

fuzzing/common.bzl
```diff
@@ -50,6 +53,10 @@ Rule for creating a script to run the fuzzing test.
             executable = True,
             cfg = "host",
         ),
+        "_engine": attr.label(
```
stefanbucur, line 56: Maybe add a constraint that this label should offer the `BuildSettingInfo` provider?

fuzzing/tools/launcher.py
```diff
@@ -44,8 +48,9 @@ def main(argv):
             "\n\tpython " + __file__ + " EXECUTABLE")
 
     command_args = [argv[1]]
-    command_args.append("-max_total_time=" + str(FLAGS.timeout_secs))
-    command_args.append("-timeout=" + str(FLAGS.timeout_secs))
+    if not FLAGS.regression:
+        command_args.append("-max_total_time=" + str(FLAGS.timeout_secs))
+        command_args.append("-timeout=" + str(FLAGS.timeout_secs))
```
stefanbucur, line 53: I believe timeouts should be available for libFuzzer regression runs too.

fuzzing/common.bzl
asraa: Yep, right now it looks like `--config asan-libfuzzer` dictates whether it runs in regression mode or not, but the command name dictates whether it is in regression mode:
`bazel test <name>` vs `bazel run <name>_run`

tengpengli:
> Just to double-check, is [envoyproxy/envoy#12530](https://github.com/envoyproxy/envoy/pull/12530) on hold until this PR is sorted out?

Yes, I plan to go back to the envoy PR as soon as this PR is merged. Can you take a look at the [comment question](https://github.com/envoyproxy/envoy/pull/12530#issuecomment-671373173) under the PR?

.bazelrc
```diff
@@ -21,12 +21,15 @@ build:clang --action_env=CXX=clang++
 build:asan-libfuzzer --config=clang
 build:asan-libfuzzer --linkopt=-fsanitize=fuzzer,address
 build:asan-libfuzzer --copt=-fsanitize=fuzzer,address
+build:asan-libfuzzer --//fuzzing:engine=libfuzzer
 
 # Flags for Clang with MSAN and libfuzzer
 build:msan-libfuzzer --config=clang
 build:msan-libfuzzer --linkopt=-fsanitize=memory,fuzzer
 build:msan-libfuzzer --copt=-fsanitize=memory,fuzzer
+build:asan-libfuzzer --//fuzzing:engine=libfuzzer
```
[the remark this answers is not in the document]
tengpengli: I think so. How do you think? @stefanbucur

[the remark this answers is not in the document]
stefanbucur: Yes, the `need_launcher` approach feels quite convoluted IMO because it's part of a rule that is actually called a launcher.

stefanbucur, line 30: Typo: Should be `msan-libfuzzer`.

The fact that no test fails because of this is a sign we don't have enough testing :) Feel free to file an issue for this.

.bazelrc
```diff
@@ -21,12 +21,15 @@ build:clang --action_env=CXX=clang++
...
 
 # Flags for Clang with MSAN and libfuzzer, outputting detailed report
 build:msan-libfuzzer-repro --config=msan-libfuzzer
 build:msan-libfuzzer-repro --copt=-fsanitize-memory-track-origins=2
+build:asan-libfuzzer --//fuzzing:engine=libfuzzer
```
stefanbucur, line 35: Why is this duplicated here?

.bazelrc
tengpengli: Created #61 for this issue.

fuzzing/common.bzl
```diff
@@ -15,19 +15,23 @@
...
+exec {launcher_path} {target_binary_path} --corpus_dir={corpus_dir} --engine={engine_type} "$@" """
 
     script_content = script_template.format(
         launcher_path = ctx.executable._launcher.short_path,
         target_binary_path = ctx.executable.target.short_path,
         corpus_dir = ctx.file.corpus.short_path if ctx.attr.corpus else "",
+        engine_type = ctx.attr._engine[BuildSettingInfo].value,
     )
+    if ctx.attr.is_regression:
+        script_content += " --regression=True"
```
stefanbucur, line 34: This approach can trick the reader to believe the `script_template` above is actually the entire script, whereas we may be adding extra content to it. Moreover, assuming that `script_content` ends in a way we can just append an argument to it also seems a bit brittle.

Consider constructing the command line for the script in a consistent manner. For example, you can first construct a vector of arguments for the launcher command line, and flatten that to a string, then substitute that string into the script template.

fuzzing/common.bzl
```diff
@@ -15,18 +15,25 @@
...
 
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
+
 def _fuzzing_launcher_impl(ctx):
     # Generate a script to launcher the fuzzing test.
     script = ctx.actions.declare_file("%s" % ctx.label.name)
+    args = []
+
+    if ctx.attr.is_regression:
+        args.append("--regression=True")
```
stefanbucur, line 26: I believe this still doesn't fully address the consistency issue: Some args are still part of the template, and others are constructed here. To give the reader a complete picture of the command line, consider building the args completely here (starting from the launcher_path, then the target binary path, etc.)

PR #62, opened 2020-08-14
Uses pip3_import rule to install absl-py

tengpengli:
**Commit Message:**
Uses pip3_import rule to install absl-py
**Additional Description:**
In the integration with Envoy, the issue that the python module absl-py can't be found raised. This PR tries to install absl-py through pip3_import rule.
**Testing:**
Local testing and CI testing.
**Docs Changes:** N/A

	modified:   WORKSPACE
	new file:   fuzzing/requirements.txt


tengpengli:
As we discussed in https://github.com/googleinterns/bazel-rules-fuzzing/pull/48, I import all dependencies in WORKSPACE directly, which seems kind of messy here. Will it be better to create a rule to import those dependencies?

fuzzing/requirements.txt
```diff
@@ -0,0 +1 @@
+absl-py==0.9.0
```
stefanbucur, line 1: Should this be `>=`? This will make it easy for other Bazel repos that need absl-py to co-exist with this one.

WORKSPACE
```diff
@@ -26,3 +26,13 @@ py_repositories()
...
+
+load("@rules_python//python:pip.bzl", "pip3_import", "pip_repositories")
+pip_repositories()
+pip3_import(
+    name = "fuzzing_py_deps",
+    requirements = "//fuzzing:requirements.txt",
+)
+
+load("@fuzzing_py_deps//:requirements.bzl", fuzzing_py_install = "pip_install")
+fuzzing_py_install()
```
stefanbucur, line 38: Shouldn't some of these be available inside `rules_fuzzing_dependencies()`? Otherwise, developers may need to re-create this setup in their own workspace.

tengpengli: I am not sure whether understand your meaning correctly.

`rules_fuzzing_dependencies` downloads the repositories, but to load the dependencies of those external repositories, it needs to load some rules of those external repositories, which can't be done inside the `rules_fuzzing_dependencies` since `load` is only available on the top level.

I created a `dependency_imports.bzl` to do the `import` job, PTAL!

stefanbucur: > I am not sure whether understand your meaning correctly.
> 
> `rules_fuzzing_dependencies` downloads the repositories, but to load the dependencies of those external repositories, it needs to load some rules of those external repositories, which can't be done inside the `rules_fuzzing_dependencies` since `load` is only available on the top level.

So how will developers (and Envoy in particular) get to use our rules? Will they need to figure out how to load our `requirements.txt` file?

> 
> I created a `dependency_imports.bzl` to do the `import` job, PTAL!

I'm not sure this addresses my concern ("developers may need to re-create this setup in their own workspace"). If this new .bzl file is only meant to be included by our WORKSPACE, then I'd prefer to have it back inlined in the WORKSPACE, so we have a complete view of our dependencies.

tengpengli: IMO, the developer now can add snippet below to their WORKSPACE to import our dependencies.
```
load("//fuzzing:repositories.bzl", "rules_fuzzing_dependencies")
rules_fuzzing_dependencies()

load("//fuzzing:dependency_imports.bzl", "fuzzing_dependency_imports")
fuzzing_dependency_imports()

load("@fuzzing_py_deps//:requirements.bzl", fuzzing_py_install = "pip_install")
fuzzing_py_install()
```
My worry is that if they are having one same dependency as ours but in different version, will our dependencies import cause any problem?

tengpengli:
Did a test in my local repository with two version of `bazel_skylib`, version 1.0.1 is explicitly imported in WORKSPACE, the other one, version 1.0.2, is implicitly imported in our `fuzzing_dependency_imports`. Then `bazel_skylib` is loaded in a BUILD file and a `string_flag` is defined. So far it doesn't cause any problem.

But according to the official recommendation on the [transitive dependencies](https://docs.bazel.build/versions/master/external.html#transitive-dependencies), it seems we shouldn't pack those imports up.

WORKSPACE
tengpengli: Did a test in my local repository with two version of `bazel_skylib`, version 1.0.1 is explicitly imported in WORKSPACE, the other one, version 1.0.2, is implicitly imported in our `fuzzing_dependency_imports`. Then `bazel_skylib` is loaded in a BUILD file and a `string_flag` is defined. So far it doesn't cause any problem.

But according to the official recommendation on the [transitive dependencies](https://docs.bazel.build/versions/master/external.html#transitive-dependencies), it seems we shouldn't pack those imports up.

PR #65, opened 2020-08-26
Deletes is_regression parameter in fuzzing_launcher

tengpengli:
**Commit Message:**
Deletes is_regression parameter in fuzzing_launcher.
**Additional Description:**
After deleting this parameter, the user can't set the launcher for regression test directly, instead the user can set it in the command or a shell script.

This change comes from the discussion in https://github.com/envoyproxy/envoy/pull/12530#discussion_r476725983

**Testing:**
Local testing and CI testing.
**Docs Changes:** N/A

### DIFF
--- a/fuzzing/tools/launcher.py
+++ b/fuzzing/tools/launcher.py
@@ -35,7 +35,7 @@ flags.DEFINE_enum(
 
 flags.DEFINE_integer(
     "timeout_secs",
-    20,
+    0,
     "The maximum duration, in seconds, of the fuzzer run launched.",
     lower_bound=0)
 
--- a/fuzzing/cc_deps.bzl
+++ b/fuzzing/cc_deps.bzl
@@ -35,6 +35,7 @@ def cc_fuzz_test(
     Args:
         name: a unique name for this target.
         corpus: a list containing corpus files.
+        dicts: a list containing dictionaries.
         **kwargs: keyword arguments.
     """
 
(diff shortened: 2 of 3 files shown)


tengpengli:
We talked about updating the `README` in every PR, but there is one thing confusing me. The commit tag is not available until the PR is merged, how can I update the snippet which downloads bazel-rules-fuzzing in the `README` with the latest commit?

asraa:
For README, let's merge this, then merge main into the README PR and make sure it's updated with any latest commits. Then we merge README PR.

tengpengli:
SGTM!

fuzzing/tools/launcher.py
```diff
@@ -35,7 +35,7 @@
 
 flags.DEFINE_integer(
     "timeout_secs",
-    20,
+    0,
```
asraa, line 38: why 0? shouldn't regression shell script manually set to 0?

asraa: ah 0 is continuous. This aligns with our current use, but I wonder if we should guarantee that the process should terminate. @stefanbucur ?

stefanbucur: That's a good question - but I'm not sure how this blocks the current PR? Can we revert this change and worry about it later?

tengpengli: The original `timeout_secs` is set to 20, thus the user will have to set this flag explicitly to get a continuous fuzzing test.

Also, the `is_regression` flag is set to True in default, which means the remove in https://github.com/envoyproxy/envoy/pull/12530#discussion_r476725983 will make the `<name>_run` target becomes a regression test. Thus that comment is not fully resolved until it is using the new version of `bazel-rules-fuzzing` with this PR.

PR #72, opened 2020-11-05
Fix the buffer overflow example in opt build mode.

stefanbucur:
In the current implementation, the compiler optimizes away the no-op
path when built in the 'opt' configuration.


asraa (approved):
nice catch, thanks!

PR #73, opened 2020-11-05
Remove the PR template. 

stefanbucur:
I'm currently working on the project primarily from VSCode, which automatically populates PRs from commit descriptions.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,0 @@
-**Commit Message:**
-**Additional Description:**
-**Testing:**
-**Docs Changes:**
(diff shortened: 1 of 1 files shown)


asraa (commented):
Are you using VScode's GH PR extension?  I think it should be able to pull the template in https://github.com/microsoft/vscode-pull-request-github/releases/tag/0.20.0 but happy to remove and revist.

stefanbucur:
Hmm, yes, I'm using that one, and I see I'm also using v0.21.0, which should already include this functionality. But in any case, I've already made a habit of including detailed change descriptions in the commit messages, so I think it should be fine to remove this template for the time being. We can add it back later of course if you or others will need it.

PR #78, opened 2020-11-15
Revamp the README documentation of the rules.

stefanbucur:
To do so, this change bumps up the python_rules dependency, which simplifies the import of PIP rules. It also adds a new RE2 example, which showcases a "real" API and demonstrates the fuzz test's ability to discover new coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,3 @@ A fuzzing engine launcher script receives configuration through the following en
 
 * [`cc_fuzz_test`](/docs/cc-fuzzing-rules.md#cc_fuzz_test)
 * [`cc_fuzzing_engine`](/docs/cc-fuzzing-rules.md#cc_fuzzing_engine)
-
-## Disclaimer
-
-This is not an officially supported Google product.
--- a/README.md
+++ b/README.md
@@ -35,8 +35,8 @@ cc_fuzz_test(
     name = "re2_fuzz_test",
     srcs = ["re2_fuzz_test.cc"],
     deps = [
-        "@re2//:re2"
-    ]
+        "@re2",
+    ],
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -1,26 +1,103 @@
 # Bazel Rules for Fuzz Tests
 
-This repository contains [Bazel](https://bazel.build/)
-[Starlark extensions](https://docs.bazel.build/versions/master/skylark/concepts.html)
-for defining fuzz tests in Bazel projects. 
+## Overview
 
-**This is not an officially supported Google product.**
+This repository contains [Bazel](https://bazel.build/) [Starlark extensions](https://docs.bazel.build/versions/master/skylark/concepts.html) for defining fuzz tests in Bazel projects.
 
-## Requirements
+Fuzzing is an effective technique for uncovering security and stability bugs in software. Fuzzing works by invoking the code under test (e.g., a library API) with automatically generated data, and observing its execution to discover incorrect behavior, such as memory corruption or failed invariants. Covering fuzzing in detail is outside the scope of this document. Read more [here](https://github.com/google/fuzzing) about fuzzing best practices, additional examples, and other resources.
 
-To use the Bazel rules for fuzzing, your C++ toolchain should be Clang-based and configured to build the fuzz tests under several compiler instrumentation modes.
+This rule library provides support for writing *in-process* fuzz tests, which consist of a driver function that receives a generated input string and feeds it to the API under test. To make a complete fuzz test executable, the driver is linked with a fuzzing engine, which implements the test generation logic. The rule library provides out-of-the-box support for the most popular fuzzing engines (e.g., [libFuzzer](https://llvm.org/docs/LibFuzzer.html) and [Honggfuzz](https://github.com/google/honggfuzz)), and an extension mechanism to define new fuzzing engines.
 
-The most convenient way to set up your toolchain is by editing the [`.bazelrc` file](https://docs.bazel.build/versions/master/guide.html#bazelrc-the-bazel-configuration-file) of your project and grouping the build configuration options into `--config` groups. We recommend using [this setup](/.bazelrc), which you can copy and paste in your own `.bazelrc` file. The setup defines the following build modes:
+The goal of the fuzzing rules is to provide an easy-to-use interface for developers to specify, build, and run fuzz tests, without worrying about the details of each fuzzing engine. A fuzzing rule wraps a raw fuzz test executable and provides additional tools, such as the specification of a corpus and dictionary and a launcher that knows how to invoke the fuzzing engine with the appropriate set of flags.
 
-- `--config=asan-libfuzzer` builds the fuzz target in [libFuzzer](https://llvm.org/docs/LibFuzzer.html) mode, with [Address Sanitizer (ASAN)](https://clang.llvm.org/docs/AddressSanitizer.html) instrumentation.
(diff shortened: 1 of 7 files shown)


fuzzing/requirements.txt
```diff
@@ -1 +1,4 @@
-absl-py>=0.9.0
+# Python requirements for the tools supporting the fuzzing rules. These are
+# installed automatically through the WORKSPACE configuration.
+
+absl-py>=0.11.0
```
inferno-chromium, line 4: nit: >= can be risky since you never know a new version breaks stuff, but upto you, we freeze version in other places.

README.md
```diff
@@ -103,3 +146,12 @@ A fuzzing engine launcher script receives configuration through the following en
...
 | `FUZZER_ARTIFACTS_DIR`     | A path under `FUZZER_OUTPUT_ROOT` where generated crashes and other relevant artifacts should be stored. |
+
+## Rule reference
+
+* [`cc_fuzz_test`](/docs/cc-fuzzing-rules.md#cc_fuzz_test)
+* [`cc_fuzzing_engine`](/docs/cc-fuzzing-rules.md#cc_fuzzing_engine)
+
+## Disclaimer
+
+This is not an officially supported Google product.
```
inferno-chromium, line 157: Do we want this disclaimer, this will be supported, and is an important part of our fuzzing portolio ?

README.md
```diff
@@ -1,91 +1,134 @@
...
+### Configuring the .bazelrc file
+
+To make sure the fuzz tests are built with the correct instrumentation flags for each engine / instrumentation supported, we recommend using the configurations defined in this repository's [`.bazelrc` file](/.bazelrc), which you can copy and paste in your own `.bazelrc` file.
+
+Currently, the following configurations are available, based on the fuzzing engines defined in this repository:
+
+| Configuration             | Fuzzing engine | Instrumentation          |
+|---------------------------|----------------|--------------------------|
+| `--config=asan-fuzzer`    | libFuzzer      | Address Sanitizer (ASAN) |
+| `--config=msan-fuzzer`    | libFuzzer      | Address Sanitizer (ASAN) |
```
inferno-chromium, line 128: This needs to Memory Sanitizer (MSAN)

README.md
```diff
@@ -1,91 +1,134 @@
...
+| `--config=asan-honggfuzz` | Honggfuzz      | Memory Sanitizer (MSAN)  |
```
inferno-chromium, line 129: This needs to Address Sanitizer (ASAN)

README.md
```diff
@@ -1,91 +1,134 @@
...
-$ bazel run fuzz_test_run --config=asan-libfuzzer -- --timeout_secs=20
+$ bazel run -c opt --config=asan-honggfuzz //examples:re2_fuzz_test_run
 ```
 
-If you only want to run the regression test on the corpus, set `--regression`:
+The `<name>_run` target accepts a number of engine-agnostic flags. For example, the following command runs the fuzz test with an execution timeout and on a clean slate (removing any previously generated tests). Note the extra `--` separator between Bazel's own flags and the launcher flags:
 
 ```sh
-$ bazel run fuzz_test_run --config=asan-libfuzzer -- --regression=True
+$ bazel run -c opt --config=asan-libfuzzer //examples:re2_fuzz_test_run -- --clean --timeout_secs=30
```
inferno-chromium, line 87: nit: Maybe break to new line \
-- --clean --timeout_secs30
to make it more obvious ?

README.md
```diff
@@ -1,91 +1,134 @@
...
 
-- `--config=asan-libfuzzer` builds the fuzz target in [libFuzzer](https://llvm.org/docs/LibFuzzer.html) mode, with [Address Sanitizer (ASAN)](https://clang.llvm.org/docs/AddressSanitizer.html) instrumentation.
-- `--config=msan-libfuzzer` builds the fuzz target in libFuzzer mode, with [Memory Sanitizer (MSAN)](https://clang.llvm.org/docs/MemorySanitizer.html) instrumentation.
-- `--config=asan-honggfuzz` builds the fuzz target using [Honggfuzz](https://github.com/google/honggfuzz), with ASAN instrumentation.
+The rule library currently provides support for C++ fuzz tests. Support for additional languages may be added in the future.
 
-The rest of the documentation assumes the build configuration options are accessible through these names.
+## Prerequisites
+
+C++ fuzz tests require a Clang compiler. The libFuzzer engine requires at least Clang 5.0.
```
inferno-chromium, line 17: i think 5.0 might be too old, i remember something like 9.0, can update after checking with kcc or dynamic-tools.

inferno-chromium (approved):

README.md
```diff
@@ -1,91 +1,134 @@
 # Bazel Rules for Fuzz Tests
 
-This repository contains [Bazel](https://bazel.build/)
-[Starlark extensions](https://docs.bazel.build/versions/master/skylark/concepts.html)
-for defining fuzz tests in Bazel projects. 
+## Overview
 
-**This is not an officially supported Google product.**
+This repository contains [Bazel](https://bazel.build/) [Starlark extensions](https://docs.bazel.build/versions/master/skylark/concepts.html) for defining fuzz tests in Bazel projects.
```
oliverchang, line 5: nit: I feel overall that this top-level README is a bit too wordy. Perhaps would be better to move parts of this to links to more detailed places? 

Ideally the README should just be a few lines showing a very simple example on how to get started very quickly. 

For more detailed information, might be better to move these other sections to sub-pages, especially sections like "Defining fuzzing engines" which should be less common but take up a lot of space here.

oliverchang (commented):
The content looks great overall, but just a comment on possibly splitting this up. I think we should make the top-level README as short and approachable as possible, and break up the more detailed sections into other files. WDYT?

README.md
stefanbucur: Done.

README.md
stefanbucur: Great catch, thanks!

README.md
stefanbucur: Done.

README.md
stefanbucur: I am open to structuring this documentation differently. The reason I went with this approach is that this seems to be the prevailing form of documentation for Bazel rules. See for example:

* https://github.com/bazelbuild/rules_python
* https://github.com/bazelbuild/rules_go
* https://github.com/bazelbuild/rules_kotlin

And there are others like that, too.

I don't feel too strongly about this, so let me know if you prefer to take a different approach here.

README.md
stefanbucur: Good catch. LibFuzzer's [documentation](https://llvm.org/docs/LibFuzzer.html#fuzzer-usage) actually mentions Clang 6.0 as being the minimum version:

> Recent versions of Clang (starting from 6.0) include libFuzzer, and no extra installation is necessary.

Updated to 6.0, but let me know if there are specific libFuzzer features that we should require and that are available in more recent versions.

stefanbucur (commented):
PTAL.

stefanbucur:
@inferno-chromium @oliverchang When you get the chance, please let me know if my latest comments address your concerns. Thanks a lot for your time reviewing this!

README.md
inferno-chromium: ok fine then.

oliverchang (approved):

README.md
inferno-chromium: As long as we have someone maintaining it, fixing bugs, then it is not needed. I havent seen it used often, e.g. https://github.com/dvyukov/go-fuzz
Upto you. Adding it decreases credibility slightly as well.

inferno-chromium:
LGTM

README.md
stefanbucur: Good point. I'm also seeing now that other Bazel rules repositories don't mention that either, so I'm removing it from here as well.

PR #82, opened 2020-11-25
Update the repo HTTP archive source location in the README example.

stefanbucur:
The commit ID is past the fix to the issue that prevented the rules from being imported in another workspace.


asraa (approved):
Thanks!

PR #94, opened 2020-12-13
Consolidate all fuzz test info (corpus, dictionaries, etc.) in one rule.

stefanbucur:
This is an internal refactoring that largely has no effect on the interface of the `cc_fuzz_test` macro.

The new `fuzzing_binary` rule supersedes the `instrumented_fuzzing_binary` rule. In addition to providing binary instrumentation, it collects all the other metadata about the fuzz test, such as corpus and dictionary paths, and the info about the fuzzing engine used, and makes it available to consumers via the new `CcFuzzingBinaryInfo` provider. This simplifies the creation of additional rules that depend on fuzz test binaries, such as the upcoming OSS-Fuzz packaging rule.

There is one major change in behavior for `cc_fuzz_test`: it now always builds the corpus and dictionary dependencies as part of building the main fuzzing binary. This is advantageous, but at the same time it invalidates an existing "compilation failure" test which was run manually. The test is disabled for the time being - I opened #95 to track this issue.


examples/BUILD
```diff
@@ -45,16 +45,6 @@ cc_fuzz_test(
     dicts = ["dict_dir/valid.dict"],
 )
 
-# This target shows a fuzz test target with an invalid dictionary.
-# Building and runing empty_fuzz_test_with_invalid_dict should be ok,
-# but building and running empty_fuzz_test_with_invalid_dict_run are expected to get an error message
-# due to the invalid dictionary.
-cc_fuzz_test(
```
markus-kusano, line 52: Is this still being used or does it still make sense to have it around (e.g., for testing)? Mostly asking since the PR description sounded like this was mostly going to be a refactoring

fuzzing/private/binary.bzl
```diff
@@ -0,0 +1,163 @@
...
+    fields = {
+        "binary_file": "The instrumented fuzz test executable.",
+        "binary_runfiles": "The runfiles of the fuzz test executable.",
+        "corpus_dir": "The directory of the corpus files used as input seeds.",
+        "dictionary_file": "The dictionary file to use in fuzzing runs.",
+        "engine_info": "The `CcFuzzingEngineInfo` provider of the fuzzing engine used in the fuzz test.",
+    },
+)
+
+def _fuzzing_binary_transition_impl(settings, attr):
```
markus-kusano, line 42: is `attr` ever used?

markus-kusano (approved):
only nits

examples/BUILD
stefanbucur: That's a great point - this was intended to be a refactoring, but it does indeed slightly change the semantics of the `cc_fuzz_test` rule: while the rule output stays the same, there are now more targets in the macro that are built by default. This test was assuming one of these targets (the dictionary) would not be built by default, so someone could manually check that this fuzz test would fail.

But since these manual "compilation failure" tests are rather brittle and less useful, I opted to remove this one for the time being.

I opened #95 to track the addition of a proper regression test.

fuzzing/private/binary.bzl
stefanbucur: It is not for this particular transition, but this is part of the standard signature required by the `implementation` attribute of the `transition()` definition below. (It's a bit like `argc` and `argv` in a main function.)

PR #96, opened 2020-12-16
Add OSS-Fuzz support to the Bazel fuzzing rules.

stefanbucur:
The support includes (a) a repository rule that extracts OSS-Fuzz environment information and makes it available hermetically in a `@rules_fuzzing_oss_fuzz` repo, and (b) a new target in the `cc_fuzz_test` macro that packages a fuzz test and its dependencies in the OSS-Fuzz format.

### DIFF
--- a/fuzzing/private/oss_fuzz/BUILD.tpl
+++ b/fuzzing/private/oss_fuzz/BUILD.tpl
@@ -1,0 +1,34 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_fuzzing//fuzzing:cc_deps.bzl", "cc_fuzzing_engine")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_fuzzing_engine(
+    name = "oss_fuzz_engine",
+    display_name = "OSS-Fuzz",
+    launcher = "oss_fuzz_launcher.sh",
+    library = ":oss_fuzz_stub",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "oss_fuzz_stub",
+    srcs = [%{stub_srcs}],
+    linkopts = [%{stub_linkopts}],
+)
+
+exports_files([
+    "instrum.bzl",
+])
--- a/fuzzing/private/oss_fuzz/package.bzl
+++ b/fuzzing/private/oss_fuzz/package.bzl
@@ -1,0 +1,71 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rule for packaging fuzz tests in the expected OSS-Fuzz format."""
+
+load("//fuzzing/private:binary.bzl", "CcFuzzingBinaryInfo")
+
+def _oss_fuzz_package_impl(ctx):
+    output_archive = ctx.actions.declare_file(ctx.label.name + ".tar")
+    binary_info = ctx.attr.binary[CcFuzzingBinaryInfo]
+
+    action_inputs = [binary_info.binary_file]
+    if binary_info.corpus_dir:
+        action_inputs.append(binary_info.corpus_dir)
+    if binary_info.dictionary_file:
+        action_inputs.append(binary_info.dictionary_file)
+    ctx.actions.run_shell(
+        outputs = [output_archive],
+        inputs = action_inputs,
+        command = """
+            declare -r STAGING_DIR="$(pwd)/{output}.staging"
+            mkdir "$STAGING_DIR"
+            ln -s "$(pwd)/{binary_path}" "$STAGING_DIR/{base_name}"
+            if [[ -n "{corpus_dir}" ]]; then
+                pushd "{corpus_dir}" >/dev/null
+                zip --quiet -r "$STAGING_DIR/{base_name}_seed_corpus.zip" ./*
+                popd >/dev/null
+            fi
+            if [[ -n "{dictionary_path}" ]]; then
+                ln -s "$(pwd)/{dictionary_path}" "$STAGING_DIR/{base_name}.dict"
+            fi
+            tar -chf "{output}" -C "$STAGING_DIR" .
+        """.format(
+            base_name = ctx.executable.binary.basename,
+            binary_path = binary_info.binary_file.path,
+            corpus_dir = binary_info.corpus_dir.path if binary_info.corpus_dir else "",
+            dictionary_path = binary_info.dictionary_file.path if binary_info.dictionary_file else "",
+            output = output_archive.path,
+        ),
+    )
+    return [DefaultInfo(files = depset([output_archive]))]
+
+oss_fuzz_package = rule(
+    implementation = _oss_fuzz_package_impl,
+    doc = """
+Packages a fuzz test in a TAR archive compatible with the OSS-Fuzz format.
+
+> NOTE: The current implementation does not yet support packaging the
+> binary runfiles.
+""",
+    attrs = {
+        "binary": attr.label(
+            executable = True,
+            doc = "The fuzz test executable.",
+            providers = [CcFuzzingBinaryInfo],
+            mandatory = True,
+            cfg = "target",
+        ),
+    },
+)
--- a/fuzzing/private/oss_fuzz/repository.bzl
+++ b/fuzzing/private/oss_fuzz/repository.bzl
@@ -1,0 +1,110 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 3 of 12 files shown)


stefanbucur:
I still need to update the documentation, but I wanted to first validate with you the implementation, to make sure it properly observes OSS-Fuzz's contracts @inferno-chromium @oliverchang

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -0,0 +1,71 @@
...
+            if [[ -n "{corpus_dir}" ]]; then
+                pushd "{corpus_dir}" >/dev/null
+                zip --quiet -r "$STAGING_DIR/{base_name}_seed_corpus.zip" ./*
+                popd >/dev/null
+            fi
+            if [[ -n "{dictionary_path}" ]]; then
+                ln -s "$(pwd)/{dictionary_path}" "$STAGING_DIR/{base_name}.dict"
+            fi
+            tar -chf "{output}" -C "$STAGING_DIR" .
+        """.format(
```
stefanbucur, line 44: @inferno-chromium @oliverchang Let me know if you notice anything wrong with this packaging structure.

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -0,0 +1,71 @@
...
+    )
+    return [DefaultInfo(files = depset([output_archive]))]
+
+oss_fuzz_package = rule(
+    implementation = _oss_fuzz_package_impl,
+    doc = """
+Packages a fuzz test in a TAR archive compatible with the OSS-Fuzz format.
+
+> NOTE: The current implementation does not yet support packaging the
+> binary runfiles.
```
stefanbucur, line 60: @inferno-chromium @oliverchang Do you know how often runfiles are an issue with OSS-Fuzz projects using Bazel today? Asking to decide how to prioritize this feature.

fuzzing/private/oss_fuzz/repository.bzl
```diff
@@ -0,0 +1,110 @@
...
+        instrum_cxxopts.append(cxxflag)
+        if cxxflag not in stub_linkopts:
+            stub_linkopts.append(cxxflag)
+
+    return struct(
+        stub_srcs = stub_srcs,
+        stub_linkopts = stub_linkopts,
+        instrum_conlyopts = instrum_conlyopts,
+        instrum_cxxopts = instrum_cxxopts,
+    )
```
stefanbucur, line 60: @asraa The logic in this function is roughly based on Envoy's own [`build.sh` file](https://github.com/google/oss-fuzz/blob/master/projects/envoy/build.sh) in OSS-Fuzz. The main difference is that these non-hermetic operations are extracted in a separate repository rule and made available to Bazel as hermetic generated repos.

Can you take a closer look and let me know if there's anything that might be missing?

fuzzing/private/oss_fuzz/repository.bzl
```diff
@@ -0,0 +1,110 @@
...
+        elif fuzzing_engine_library.endswith(".a"):
+            repository_ctx.symlink(
+                repository_ctx.path(fuzzing_engine_library),
+                "oss_fuzz_engine.a",
+            )
+            stub_srcs.append("oss_fuzz_engine.a")
+        else:
+            fail("Unsupported $LIB_FUZZING_ENGINE value '%s'" % fuzzing_engine_library)
+    for cflag in cflags:
+        if cflag == "-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION":
```
asraa, line 43: i think these and in cxxflags should be removed (this reminds me Envoy's build can remove this flag now. we originally had it because a 3p dependency built with this flag would break our fuzzers, and it was a workaround.)

stefanbucur:
Starting another thread to discuss testing: IMO, it would be nice if there was a way to make sure the code works correctly end-to-end on the latest versions of these rules and OSS-Fuzz.

@inferno-chromium @oliverchang do you think we should have tests inside the OSS-Fuzz repo, here, or both? If so, any thoughts on how to best test these? My own (manual) testing resorted to create a dummy project (like [this](https://github.com/stefanbucur/oss-fuzz/tree/bazel-rules-fuzzing/projects/bazel-rules-fuzzing)) and then making sure the examples in this repo build under that project. Is there a better way to do this during CI?

fuzzing/private/oss_fuzz/package.bzl
oliverchang: I don't think this is an issue with the current projects. I eyeballed the projects that use bazel, and couldn't see any that copies the runfiles: https://github.com/google/oss-fuzz/search?q=bazel

fuzzing/private/oss_fuzz/package.bzl
oliverchang: The seed_corpus and dict paths look good, but would it make sense support generating options? 

e.g. letting people support specifying options in the rule that would generate something like: https://github.com/google/oss-fuzz/blob/master/projects/harfbuzz/hb-draw-fuzzer.options

fuzzing/private/oss_fuzz/BUILD.tpl
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_fuzzing//fuzzing:cc_deps.bzl", "cc_fuzzing_engine")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_fuzzing_engine(
+    name = "oss_fuzz_engine",
+    display_name = "OSS-Fuzz",
+    launcher = "oss_fuzz_launcher.sh",
```
oliverchang, line 21: What is this for?

fuzzing/private/oss_fuzz/repository.bzl
```diff
@@ -0,0 +1,110 @@
...
+    )
+    repository_ctx.file(
+        "oss_fuzz_launcher.sh",
+        "echo 'The OSS-Fuzz engine is not meant to be executed.'; exit 1",
+    )
+
+oss_fuzz_repository = repository_rule(
+    implementation = _oss_fuzz_repository,
+    environ = [
+        "LIB_FUZZING_ENGINE",
```
oliverchang, line 99: These are the env vars that are forwarded to bazel right? 

@jonathanmetzman anything missing here?

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -0,0 +1,71 @@
...
+""",
+    attrs = {
+        "binary": attr.label(
```
oliverchang, line 63: Does this mean that the oss_fuzz_package rule creates a separate tar for each target binary?

oliverchang (commented):
nice!

fuzzing/private/oss_fuzz/repository.bzl
jonathanmetzman: Don't think so.

stefanbucur: Thanks for reviewing!

stefanbucur (commented):
Thanks for all the feedback! PTAL.

stefanbucur:
> Starting another thread to discuss testing: IMO, it would be nice if there was a way to make sure the code works correctly end-to-end on the latest versions of these rules and OSS-Fuzz.
> 
> @inferno-chromium @oliverchang do you think we should have tests inside the OSS-Fuzz repo, here, or both? If so, any thoughts on how to best test these? My own (manual) testing resorted to create a dummy project (like [this](https://github.com/stefanbucur/oss-fuzz/tree/bazel-rules-fuzzing/projects/bazel-rules-fuzzing)) and then making sure the examples in this repo build under that project. Is there a better way to do this during CI?

Bumping this up, in case you have any suggestion.

fuzzing/private/oss_fuzz/package.bzl
oliverchang: There are currently bazel projects which rely on .options files being copied to $OUT. e.g. https://github.com/google/oss-fuzz/blob/c1f1afe5300dc039b8e0988b4d8c1bee77fc7599/projects/grpc/build.sh#L127

I suppose they can still do this manual step/workaround to copy manually, but it would be nice if this was supported out of the box if we want to migrate these projects.

fuzzing/private/oss_fuzz/package.bzl
oliverchang: Agreed the current way seems easier.

fuzzing/private/oss_fuzz/BUILD.tpl
oliverchang: Thanks for explaining!

oliverchang (approved):
> > Starting another thread to discuss testing: IMO, it would be nice if there was a way to make sure the code works correctly end-to-end on the latest versions of these rules and OSS-Fuzz.
> > @inferno-chromium @oliverchang do you think we should have tests inside the OSS-Fuzz repo, here, or both? If so, any thoughts on how to best test these? My own (manual) testing resorted to create a dummy project (like [this](https://github.com/stefanbucur/oss-fuzz/tree/bazel-rules-fuzzing/projects/bazel-rules-fuzzing)) and then making sure the examples in this repo build under that project. Is there a better way to do this during CI?
> 
> Bumping this up, in case you have any suggestion.

Perhaps we can have a GitHub action for this repo which checks out OSS-Fuzz at the latest revision and builds the test project? Seems like something that would fit better in this repo.

fuzzing/private/oss_fuzz/package.bzl
stefanbucur: Oh, I wasn't aware of these, thanks! I briefly scanned their options files, and they seem to set libfuzzer options such as the maximum input size. It looks like these would be useful for all the consumers of the fuzz target rule, including the launcher rule.

I filed #103 and #104 for this so I can address it more systematically and keep this PR focused on the basic OSS-Fuzz integration. This seems to be indeed a blocker for the broader OSS-Fuzz integration.

stefanbucur:
> > > Starting another thread to discuss testing: IMO, it would be nice if there was a way to make sure the code works correctly end-to-end on the latest versions of these rules and OSS-Fuzz.
> > > @inferno-chromium @oliverchang do you think we should have tests inside the OSS-Fuzz repo, here, or both? If so, any thoughts on how to best test these? My own (manual) testing resorted to create a dummy project (like [this](https://github.com/stefanbucur/oss-fuzz/tree/bazel-rules-fuzzing/projects/bazel-rules-fuzzing)) and then making sure the examples in this repo build under that project. Is there a better way to do this during CI?
> > 
> > 
> > Bumping this up, in case you have any suggestion.
> 
> Perhaps we can have a GitHub action for this repo which checks out OSS-Fuzz at the latest revision and builds the test project? Seems like something that would fit better in this repo.

Sounds great - let's move the conversation to #105. There are a few details to figure out I think.

PR #98, opened 2020-12-29
Start the work on a default replay (regression) engine used for executing a corpus or set of tests

stefanbucur:
This change imports the necessary dependencies (ABSL and gTest) and adds a small library for producing absl::Status values from errno values.


fuzzing/replay/status_util.h
```diff
@@ -0,0 +1,26 @@
...
+
+// Utilities for working with absl::Status values.
+
+#include "absl/status/status.h"
+#include "absl/strings/string_view.h"
+
+namespace fuzzing {
+
+// Creates an error status value that includes the given `message` and a
+// representation of the `errno_value`.
```
xinhaoyuan, line 26: nitnit: `representation` sounds a bit abstract, how about `description`?

also I feel that `errno` is enough instead of `errno_value`, since `errno` already implies that it is a number - leaving it to you.

fuzzing/replay/status_util_test.cc
```diff
@@ -0,0 +1,55 @@
...
+TEST(StatusUtilTest, NonemptyMessage) {
+  const absl::Status status = ErrnoStatus("could not open file", ENOENT);
+  EXPECT_EQ(status.code(), absl::StatusCode::kUnknown);
+  EXPECT_THAT(status.message(),
+              StrEq("could not open file (No such file or directory)"));
+}
+
+TEST(StatusUtilTest, SuccessfulErrno) {
+  const absl::Status status = ErrnoStatus("no error", 0);
+  EXPECT_EQ(status.code(), absl::StatusCode::kUnknown);
```
xinhaoyuan, line 43: Is it possible that some people would expect absl::StatusCode::kOk here? Maybe it's worth to explain this in the header.

xinhaoyuan (approved):
LGTM

fuzzing/replay/status_util.h
stefanbucur: > nitnit: `representation` sounds a bit abstract, how about `description`?

Done.
 
> also I feel that `errno` is enough instead of `errno_value`, since `errno` already implies that it is a number - leaving it to you.

I agree, but it turns out that `errno` is defined as a macro, at least on the glibc version on my machine, so when I try to use it as a variable name, I get weird compilation errors.

After I ended up using `errno_value`, I saw that POSIX man pages use `errnum` to refer to this value in arguments, but I didn't end up making the change here. If you think it makes a difference, I can revisit this.

fuzzing/replay/status_util_test.cc
stefanbucur: Hmm, that's a great observation. Let me actually change the implementation and make sure that a zero errno results in an OkStatus().

PR #99, opened 2021-01-05
Add a set of file-related utilities that will be used in corpus traversal during replay.

stefanbucur:


fuzzing/replay/file_util.cc
```diff
@@ -0,0 +1,98 @@
...
+    const std::string& path,
+    absl::FunctionRef<void(absl::string_view, const struct stat&)> callback) {
+  struct stat path_stat;
+  if (stat(path.c_str(), &path_stat) < 0) {
+    return ErrnoStatus(absl::StrCat("could not stat ", path), errno);
+  }
+  if (S_ISDIR(path_stat.st_mode)) {
+    return TraverseDirectory(path, callback);
+  } else {
+    callback(path, path_stat);
```
oliverchang, line 78: nit: unnecessary else after return.

fuzzing/replay/file_util.cc
```diff
@@ -0,0 +1,98 @@
...
+  FILE* f = fopen(path.c_str(), "w");
+  if (!f) {
+    return ErrnoStatus("could not open file", errno);
+  }
+  const size_t result = fwrite(contents.data(), 1, contents.size(), f);
+  fclose(f);
+  if (result < contents.size()) {
+    return absl::UnknownError("could not write file contents");
+  } else {
+    return absl::OkStatus();
```
oliverchang, line 94: nit: unnecessary else after return.

oliverchang (approved):

fuzzing/replay/file_util.cc
stefanbucur: Fixed.

fuzzing/replay/file_util.cc
stefanbucur: Fixed.

PR #107, opened 2021-01-08
Add a test rule for running a fuzz test against the seed corpus.

stefanbucur:
This is a specialization of the launcher rule, defined as a test. The cc_fuzz_test macro is also refactored such that the test rule runs by default when invoking "bazel test".

### DIFF
--- a/fuzzing/private/fuzz_test.bzl
+++ b/fuzzing/private/fuzz_test.bzl
@@ -40,6 +40,11 @@ def cc_fuzz_test(
       rarely.
     * `<name>_run`: An executable target used to launch the fuzz test using a
       simpler, engine-agnostic command line interface.
+    * `<name>_oss_fuzz`: Generates a `<name>_oss_fuzz.tar` archive containing
+      the fuzz target executable and its associated resources (corpus,
+      dictionary, etc.) in a format suitable for unpacking in the $OUT/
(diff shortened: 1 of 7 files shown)


fuzzing/private/fuzz_test.bzl
```diff
@@ -57,26 +54,23 @@ def cc_fuzz_test(
           binary rule.
     """
 
+    raw_binary_name = name + "_raw_"
```
xinhaoyuan, line 57: it is intended to be `_raw_`?

fuzzing/private/fuzz_test.bzl
```diff
@@ -91,15 +85,20 @@ def cc_fuzz_test(
...
+    fuzzing_regression_test(
+        name = name,
+        binary = instrum_binary_name,
+        tags = (tags or []) + [
+            "fuzz-test",
+        ],
     )
 
     oss_fuzz_package(
         name = name + "_oss_fuzz",
```
xinhaoyuan, line 101: this may be unrelated - should we document this rule?

xinhaoyuan (approved):

fuzzing/private/fuzz_test.bzl
stefanbucur: Great catch, done!

PR #108, opened 2021-01-08
Define a buffer class suitable for sequentially reading a potentially large set of corpus files.

stefanbucur:


fuzzing/replay/test_file_buffer.h
```diff
@@ -0,0 +1,71 @@
...
+// A memory buffer for reading and storing test data from files.
+//
+// This class permits the use of a single buffer to store the data from multiple
+// test files read sequentially, resulting in reduced memory churn for large
+// corpora consisting of multiple files.
+class TestFileBuffer {
+ public:
+  // Creates a new buffer holding up to `max_size` bytes.
+  TestFileBuffer(size_t max_size)
+      : max_size_(max_size), last_size_(0), buffer_(new char[max_size]) {
```
xinhaoyuan, line 37: nit: with negative max_size, new char[max_size] will throw an exception with minimal information. Perhaps we can initialize it later than the assert?

xinhaoyuan (approved):

fuzzing/replay/test_file_buffer.h
stefanbucur: Oh that's a great point - fixed!

PR #110, opened 2021-01-12
Separate the set of associated targets in a cc_fuzz_test rule in a distinct macro.

stefanbucur:
The macro can be instantiated in custom definitions of the cc_fuzz_test macro.

### DIFF
--- a/fuzzing/private/fuzz_test.bzl
+++ b/fuzzing/private/fuzz_test.bzl
@@ -29,7 +29,7 @@ def fuzzing_decoration(
         tags = None):
     """Generates the standard targets associated to a fuzz test.
 
-    This target can be used to define custom fuzz test rules in case the default
+    This macro can be used to define custom fuzz test rules in case the default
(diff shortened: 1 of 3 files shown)


fuzzing/private/fuzz_test.bzl
```diff
@@ -20,6 +20,75 @@ load("//fuzzing/private:binary.bzl", "fuzzing_binary")
...
+def fuzzing_decoration(
+        base_name,
+        raw_binary,
+        engine,
+        corpus = None,
+        dicts = None,
+        tags = None):
+    """Generates the standard targets associated to a fuzz test.
+
+    This target can be used to define custom fuzz test rules in case the default
```
xinhaoyuan, line 32: maybe This target -> This macro?

xinhaoyuan (approved):

fuzzing/private/fuzz_test.bzl
stefanbucur: Good point - done.

PR #112, opened 2021-01-13
Pin the compiler version used in this workspace and for CI workflows.

stefanbucur:


.bazelrc
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 # Force the use of Clang for all builds.
-build --action_env=CC=clang
-build --action_env=CXX=clang++
+build --action_env=CC=clang-10
```
oliverchang, line 16: Does this make these rules less portable? i.e. on distros that use a different clang version. 

Or is this test/CI only?

oliverchang (approved):
LGTM with a question

.bazelrc
stefanbucur: Ah, this only applies for this repository's `WORKSPACE`, which is used for rule development, CI, and testing. The projects using these rules have their own `WORKSPACE` files and they set up their own C++ toolchain.

PR #116, opened 2021-01-21
Fix some missing dependencies.

xinhaoyuan:


stefanbucur (approved):
Nice catch, thanks!

BTW, how did you discover these missing dependencies? Did you find a tool to do this for you? I was trying to find one myself, too :)

xinhaoyuan:
> Nice catch, thanks!
> 
> BTW, how did you discover these missing dependencies? Did you find a tool to do this for you? I was trying to find one myself, too :)

I enabled the layering check feature which enforces that any header included in a source of a target must be in the target or a direct dependency of the target. I also fixed more missing dependencies by checking that into .bazelrc. PTAL

stefanbucur:
LGTM, thanks!

PR #119, opened 2021-01-26
Implement a number of changes needed for Envoy integration.

stefanbucur:
These changes stem from limitations in Bazel and RBE that cause transitions to balloon the size of the Envoy build output from ~9 GiB to ~28 GiB. Specifically, this change:

* Allows the fuzz targets to be instrumented outside the fuzz test macro, essentially providing a mode in which the instrumentation transition is disabled.
* Tags all non-test rules as "manual". This is needed, such that `bazel test //...` commands don't inadvertently try to build the runner tool, OSS-Fuzz package, etc, which would cause RBE to download all those artifacts locally.
* Allow disabling regression tests, allowing the caller to define their own regression (in the case of Envoy, by using the `cc_test` itself). This is also due to RBE behavior that causes the regression test rule depending on binary rules to fetch more binary artifacts locally. This behavior needs to be investigated more carefully.


docs/cc-fuzzing-rules.md
```diff
@@ -93,6 +94,8 @@ documentation for the set of targets generated.
 | <a id="fuzzing_decoration-engine"></a>engine |  The label of the fuzzing engine used to build the binary.   |  none |
 | <a id="fuzzing_decoration-corpus"></a>corpus |  A list of corpus files.   |  <code>None</code> |
 | <a id="fuzzing_decoration-dicts"></a>dicts |  A list of fuzzing dictionary files.   |  <code>None</code> |
-| <a id="fuzzing_decoration-tags"></a>tags |  Tags set on the fuzzing regression test.   |  <code>None</code> |
+| <a id="fuzzing_decoration-instrument_binary"></a>instrument_binary |  If true, depend on the <code>raw_binary</code> through a Bazel   configuration transition that uses configuration flags from the   <code>@rules_fuzzing//fuzzing</code> package to determine fuzzing build mode,   engine, and sanitizer instrumentation. If false, assumes that   <code>raw_binary</code> is already built in the proper configuration.   |  <code>True</code> |
```
asraa, line 97: Should this mention that users probably don't need to use this (and/or link to somewhere that they can understand why this is there?)

stefanbucur: You're right - I think we can go even further and just ask users to reach out to us if they think the default instrumentation mode doesn't work for them. I clarified this in the docs.

fuzzing/private/fuzz_test.bzl
```diff
@@ -41,28 +43,59 @@ def fuzzing_decoration(
         engine: The label of the fuzzing engine used to build the binary.
         corpus: A list of corpus files.
         dicts: A list of fuzzing dictionary files.
-        tags: Tags set on the fuzzing regression test.
+        instrument_binary: **(Experimental, may be removed in the future.)**
```
xinhaoyuan, line 46: if there is a plan, maybe explain what would be the behavior of the rule without this flag?

xinhaoyuan (approved):

fuzzing/private/fuzz_test.bzl
stefanbucur: Good point - I've clarified what the default behavior should be here. I'm going to assume that there is an implicit understanding that removing the flag will preserve the default behavior.

PR #121, opened 2021-01-29
Pin the hashes of all Python requirements.

stefanbucur:
This helps the fuzzing rules be consistent with Envoy's dependency policies (envoyproxy/envoy#14834).


inferno-chromium (approved):

stefanbucur:
Thanks for the quick review @inferno-chromium !

PR #125, opened 2021-02-24
Fix dictionary concatenation bug by enforcing newlines on each dictionary entry.

### LINKED ISSUE #124: Dictionary concatenation
## Expected Behavior
Expect that two dictionary files 
```
a.dict
# Dictionary A
abc
```
and
```
b.dict
# Dictionary B
def
```
are successfully concatenated together.

## Actual Behavior
It seems like if file A doesn't have an end new line, then https://github.com/bazelbuild/rules_fuzzing/blob/547127916d662d729ed62164bb88edbff27e215b/fuzzing/tools/validate_dict.py#L43 won't write one, so the resulting dictionary becomes

```
# Dictionary A
abc# Dictionary B
def
```
which isn't valid.

I originally saw the issue in oss-fuzz logs:
```
Step #14: ParseDictionaryFile: error in line 24
Step #14: 		"%HOSTNAME%"# Dictionary to populate HeaderValueOptions
```

stefanbucur:
Fixes #124


asraa (approved):
LGTM! It looks like libfuzzer strips whitespace before running the rest of the validations, so this makes sense to me
https://github.com/rust-fuzz/libfuzzer-sys/blob/35ce7d7177c254b9c798aec171dfe76877d1a20f/libfuzzer/FuzzerUtil.cpp#L85

PR #130, opened 2021-03-15
Force the linking of the UBSAN runtime in OSS-Fuzz builds.

### LINKED ISSUE #129: UBSan fails in OSS-Fuzz mode
We probably need to update the repository rule to incorporate the `--linkopt="$(find $(llvm-config --libdir) -name libclang_rt.ubsan_standalone_cxx-x86_64.a | head -1)"` argument.

stefanbucur:
Fixes #129

### DIFF
--- a/fuzzing/private/oss_fuzz/repository.bzl
+++ b/fuzzing/private/oss_fuzz/repository.bzl
@@ -14,18 +14,50 @@
 
 """Repository rule for configuring the OSS-Fuzz engine and instrumentation."""
 
+_UBSAN_STANDALONE_CXX_LIB_NAME = "libclang_rt.ubsan_standalone_cxx-x86_64.a"
+
 def _to_list_repr(elements):
     return ", ".join([repr(element) for element in elements])
+
+def _find_llvm_lib(repository_ctx, target_file):
+    result = repository_ctx.execute([
+        repository_ctx.which("bash"),
+        "-c",
+        """
+        set -euf -o pipefail
+        set -x
+        find "$({llvm_config} --libdir)" -name {target_file} | head -1
+        """.format(
+            llvm_config = "llvm-config-9",
+            target_file = target_file,
+        ),
+    ], quiet = False)
+    file_path = result.stdout.strip()
+
+    if result.return_code != 0 or not file_path:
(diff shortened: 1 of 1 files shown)


fuzzing/private/oss_fuzz/repository.bzl
```diff
@@ -14,19 +14,51 @@
...
+        """
+        set -euf -o pipefail
+        set -x
+        find "$({llvm_config} --libdir)" -name {target_file} | head -1
+        """.format(
+            llvm_config = "llvm-config",
+            target_file = target_file,
+        ),
+    ], quiet = False)
+    file_path = result.stdout.strip()
```
inferno-chromium, line 35: Can result.stdout be None in any case, that can cause an exception.

fuzzing/private/oss_fuzz/repository.bzl
```diff
@@ -14,19 +14,51 @@
...
         fuzzing_engine_library,
+        sanitizer,
         cflags,
         cxxflags):
     stub_srcs = []
     stub_linkopts = []
     instrum_conlyopts = []
     instrum_cxxopts = []
 
+    if sanitizer == "undefined":
```
inferno-chromium, line 52: Do we need to check arch is x64 or should we support both via sustituting the arch in filename (can skip for now, can add todo for later too if needed).

inferno-chromium (approved):

fuzzing/private/oss_fuzz/repository.bzl
stefanbucur: I don't think this can be None. According to [the docs](https://docs.bazel.build/versions/master/skylark/lib/exec_result.html), this is always set to the stdout contents of the program (i.e., an empty string if there's no output).

inferno-chromium: sg, skip!

fuzzing/private/oss_fuzz/repository.bzl
stefanbucur: That's a good point, it's actually easy to parameterize this by the architecture. I've just done so.

PR #137, opened 2021-04-07
Add an optional flag for listing corpus paths in lines instead of sep…

xinhaoyuan:
…arated by comma. This is to support corpus paths that contain commas.

### DIFF
--- a/fuzzing/private/common.bzl
+++ b/fuzzing/private/common.bzl
@@ -79,15 +79,19 @@ Rule for creating a script to run the fuzzing test.
 )
 
 def _fuzzing_corpus_impl(ctx):
+    corpus_list_file_args = ctx.actions.args()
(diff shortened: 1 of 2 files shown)


fuzzing/private/common.bzl
```diff
@@ -79,15 +79,19 @@ Rule for creating a script to run the fuzzing test.
 )
 
 def _fuzzing_corpus_impl(ctx):
+    corpus_list_file_args = ctx.actions.args()
+    corpus_list_file_args.set_param_file_format("multiline")
+    corpus_list_file_args.use_param_file("--corpus_list_file=%s", use_always = True)
```
stefanbucur, line 84: TIL. Neat!

stefanbucur (approved):

Issue #138, opened 2021-04-12
Build fails under OSX


eustas:
## Expected Behavior
Instructions given in README should lead to successful build.

## Actual Behavior
Build fails on OSX:
```
ERROR: /private/var/tmp/_bazel_eustas/a0a6f697b808c7d2a5fc66d10631ebe1/external/rules_fuzzing/fuzzing/replay/BUILD:48:11: Compiling fuzzing/replay/status_util.cc failed: (Exit 1): wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG '-std=c++11' ... (remaining 28 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG '-std=c++11' ... (remaining 28 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
external/rules_fuzzing/fuzzing/replay/status_util.cc:40:10: error: no viable conversion from returned value of type 'int' to function return type 'std::string' (aka 'basic_string<char, char_traits<char>, allocator<char> >')
  return strerror_r(errno_value, error_str_buf, sizeof(error_str_buf));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string:799:5: note: candidate constructor not viable: no known conversion from 'int' to 'const std::__1::basic_string<char> &' for 1st argument
    basic_string(const basic_string& __str);
    ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string:804:5: note: candidate constructor not viable: no known conversion from 'int' to 'std::__1::basic_string<char> &&' for 1st argument
    basic_string(basic_string&& __str)
    ^
[...]

PR #139, opened 2021-04-12
Fix OSX build

eustas:
* more portable strerror_r handling (borrowed from abseil)
 * adjust honggfuzz build options
 * notify linker about weak ref in :replay_main

See https://github.com/bazelbuild/rules_fuzzing/issues/138

### DIFF
--- a/fuzzing/replay/status_util.cc
+++ b/fuzzing/replay/status_util.cc
@@ -14,8 +14,10 @@
 
 #include "fuzzing/replay/status_util.h"
 
+#include <cstdio>
 #include <cstring>
 #include <string>
+#include <type_traits>
 
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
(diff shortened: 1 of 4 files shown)


fuzzing/replay/status_util.cc
```diff
@@ -26,19 +26,23 @@ namespace {
...
   if (result) {
     return absl::StrCat("Unknown error ", errno_value);
   } else {
-    return error_str_buf;
+    return WrapError(error_str_buf, errno_value, error_str_buf);
   }
-#else
-  return strerror_r(errno_value, error_str_buf, sizeof(error_str_buf));
-#endif
+}
```
stefanbucur, line 39: Since we're changing this anyway, perhaps let's adopt the logic used in ABSL for consistency? It's also known to work across more platforms. https://github.com/abseil/abseil-cpp/blob/3b4a16abad2c2ddc494371cc39a2946e36d35d11/absl/base/internal/strerror.cc#L32-L51

Ideally, we could have directly included that code, alas its build target has restricted visibility.

stefanbucur (changes requested):
The fuzzing rules were not tested on MacOS, thanks for making the code more portable.

fuzzing/replay/status_util.cc
eustas: Done.
Agree. Perhaps we should file an issue to make this a part of public API?

eustas:
While project builds now, all "examples/" currently fail

eustas:
```
Executing tests from //examples:empty_fuzz_test
-----------------------------------------------------------------------------
readlink: illegal option -- f
usage: readlink [-n] [file ...]
/private/var/tmp/_bazel_eustas/ef657362fb12bedcb3f55b48714eb83e/sandbox/darwin-sandbox/1798/execroot/rules_fuzzing/bazel-out/darwin-opt/bin/examples/empty_fuzz_test.runfiles/rules_fuzzing/fuzzing/engines/replay_launcher.sh: line 24: exec: : not found
```

eustas:
Using `realpath` instead of `readlink -f` seems to work. Will address that in next MR, alongside with adding osx-github-action.

stefanbucur (approved):

PR #140, opened 2021-04-15
Improve dict validation

### LINKED ISSUE #47: Replace the dictionary entry validation logic with a regular expression
## Expected Behavior
Use a regular expression to validate the dictionary entry in dict_validator.py.

## Actual Behavior
The function is using `if-else` logic to process each line of the dictionary.

xinhaoyuan:
Refactor dict validation using regular expression for simplification and improvement.
After the change, the validation will:

 - Disallow `"\"`, `""`, and any non-space suffix after the entry ends (e.g. `"entry"suffix`).
 - Disallow unesacped '"' in enties. Raw '"' can indeed be used in libFuzzer due to implementation details,
   but it is undocumented (https://llvm.org/docs/LibFuzzer.html#dictionaries), implementation-dependent, and confusing.
 - Show file paths and line numbers with errors.

Fixes #47

### DIFF
--- a/fuzzing/tools/dict_validation.py
+++ b/fuzzing/tools/dict_validation.py
@@ -17,47 +17,22 @@
 Validates the fuzzing dictionary.
 """
 
-from string import hexdigits
+import re
 
-
-def validate_entry(entry):
-    """Validates a single fuzzing dictionary entry.
-
-    Args:
-        entry: a string containing a single entry.
-
-    Returns:
-        True if the argument is a valid fuzzing dictionary entry, 
-        otherwise False.
-    """
-
-    # Use set to contain hex digits to decrease the query time complexity
-    hex_set = set(hexdigits)
-    pos, end = 0, len(entry) - 1
-    while pos < end:
-        pos += 1
-        chr = entry[pos]
-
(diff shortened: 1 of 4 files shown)


fuzzing/tools/dict_validation.py
```diff
@@ -43,18 +43,20 @@ def validate_entry(entry):
...
+            if pos + 3 < end and entry[pos + 1] == 'x' and entry[
                     pos + 2] in hex_set and entry[pos + 3] in hex_set:
                 pos += 3
                 continue
 
             return False
+        elif chr == '"' and pos < end:
+            return False
 
     return True
```
stefanbucur, line 61: There were several recent fixes to this code. This makes me think we should consider switching this to a more compact and easier to review regular expression. WDYT?

stefanbucur (changes requested):

fuzzing/tools/dict_validation.py
xinhaoyuan: That's a good point. Using regular expressions indeed makes it simpler and easier to understand. PTAL

stefanbucur (approved):

PR #144, opened 2021-05-02
Decompose FuzzingEngineInfo into orthogonal providers

fmeum:
By removing the engine_library_info field of CcFuzzingEngineInfo and relying exlusively on the DefaultInfo and language-specific *Info providers to propagate the information about the engine library, CcFuzzingEngineInfo can become almost completely generic in the language.

This PR performs the required refactoring and renames the provider to FuzzingEngineInfo. It also drops the Cc prefix of CcFuzzingBinaryInfo, which no longer depends on the language.

### DIFF
--- a/fuzzing/private/engine.bzl
+++ b/fuzzing/private/engine.bzl
@@ -14,20 +14,19 @@
 
 """The implementation of the cc_fuzzing_engine rule."""
 
-CcFuzzingEngineInfo = provider(
+FuzzingEngineInfo = provider(
     doc = """
 Provider for storing the specification of a fuzzing engine.
 """,
     fields = {
         "display_name": "A string representing the human-readable name of the fuzzing engine.",
-        "engine_library_info": "A CcInfo provider for the C++ library of the fuzzing engine.",
         "launcher": "A file representing the shell script that launches the fuzz target.",
         "launcher_runfiles": "The runfiles needed by the launcher script on the fuzzing engine side, such as helper tools and their data dependencies.",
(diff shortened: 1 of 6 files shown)


fmeum:
I added a new commit which finally makes #143 consist of new features only and no refactorings.

fuzzing/private/engine.bzl
```diff
@@ -44,15 +43,20 @@ def _cc_fuzzing_engine_impl(ctx):
...
         launcher = ctx.file.launcher,
         launcher_runfiles = launcher_runfiles,
         launcher_environment = env_vars,
     )
-    return [ctx.attr.library[DefaultInfo], cc_library_info, cc_fuzzing_engine_info]
+    return [
+        ctx.attr.library[DefaultInfo],
+        ctx.attr.library[language_provider],
+        fuzzing_engine_info,
+    ]
```
stefanbucur, line 56: Maybe we can push the return of this vector in the top-level _impl function? This way, we can achieve a better separation of concerns and have this function only compute and return the FuzzingEngineInfo provider (could call it, e.g., `_get_fuzzing_engine_info`), with no need to forward the language-specific provider.

stefanbucur (changes requested):
Mostly looks good, just a small change.

fuzzing/private/engine.bzl
fmeum: Right, that's better. I pushed new commits that do this.

stefanbucur (approved):

PR #145, opened 2021-05-07
Prepare Java release

fmeum:
This updates Jazzer and adds Java documentation to the README and Stardoc.

@stefanbucur I have some ideas on how to implement oss-fuzz support, but I don't think this is critical for the next release - most OSS Java projects don't use Bazel.

### DIFF
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -69,3 +69,33 @@ stardoc(
         ":cc_fuzzing_rules",
     ],
 )
+
+bzl_library(
+    name = "java_fuzzing_rules",
+    srcs = [
+        "//fuzzing:instrum_opts.bzl",
+        "//fuzzing:java_defs.bzl",
+        "//fuzzing/private:binary.bzl",
+        "//fuzzing/private:common.bzl",
+        "//fuzzing/private:engine.bzl",
+        "//fuzzing/private:fuzz_test.bzl",
+        "//fuzzing/private:instrum_opts.bzl",
+        "//fuzzing/private:java_utils.bzl",
+        "//fuzzing/private:regression.bzl",
+        "//fuzzing/private/oss_fuzz:package.bzl",
+        "@rules_fuzzing_oss_fuzz//:instrum.bzl",
+    ],
+    deps = [
+        ":bazel_skylib",
+        ":rules_cc",
+    ],
+)
+
+stardoc(
+    name = "java_fuzzing_docs",
+    out = "java-fuzzing-rules.md",
+    input = "//fuzzing:java_defs.bzl",
+    deps = [
+        ":java_fuzzing_rules",
+    ],
+)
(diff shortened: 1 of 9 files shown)


fmeum:
I think I know how to implement OSS-Fuzz support in a relatively simple way now. Maybe it does make sense to wait for it.

fmeum:
I added three commits with small Java-related fixes or improvements.

docs/BUILD
```diff
@@ -69,3 +69,33 @@ stardoc(
...
+        "//fuzzing:java_defs.bzl",
+        "//fuzzing/private:binary.bzl",
+        "//fuzzing/private:common.bzl",
+        "//fuzzing/private:engine.bzl",
+        "//fuzzing/private:fuzz_test.bzl",
+        "//fuzzing/private:instrum_opts.bzl",
+        "//fuzzing/private:java_utils.bzl",
+        "//fuzzing/private:regression.bzl",
+        "//fuzzing/private/oss_fuzz:package.bzl",
+        "@rules_fuzzing_oss_fuzz//:instrum.bzl",
```
stefanbucur, line 86: Should we just merge `java_fuzzing_rules` and `cc_fuzzing_rules` together? We can then list them in both `cc_fuzzing_docs` and `java_fuzzing_docs`.

examples/java/com/example/JavaNativeFuzzTest.cpp
```diff
@@ -16,15 +16,16 @@
 
 #include <string>
 
-// simple function containing a crash that requires coverage and string compare
-// instrumentation for the fuzzer to find
+// Prevent the compiler from optimizing away the out-of-bounds read in
+// parseInternal.
+volatile int dummy = 0;
+
```
stefanbucur, line 22: Maybe we can achieve the same result by marking the function below as `__attribute__((optnone))`?

examples/java/com/example/JavaNativeFuzzTest.cpp
```diff
@@ -16,15 +16,16 @@
...
 void parseInternal(const std::string &input) {
```
stefanbucur, line 23:
```suggestion
__attribute__((optnone)) void parseInternal(const std::string &input) {
```

stefanbucur (commented):
Mostly LGTM, a few small comments.

examples/java/com/example/JavaNativeFuzzTest.cpp
fmeum: Thanks, I didn't know this attribute.

fmeum: Thanks, I didn't know this attribute.

fmeum:
Thanks for the review! I pushed a new commit addressing the fixes.

stefanbucur (approved):
Thank you!

PR #146, opened 2021-05-10
Add OSS-Fuzz support for `java_fuzz_test`

fmeum:
This is based on #145, which should be reviewed first.

### DIFF
--- a/fuzzing/private/java_utils.bzl
+++ b/fuzzing/private/java_utils.bzl
@@ -90,16 +90,28 @@ def _java_segment_index(path_segments):
 
     return root_index
 
-def _jazzer_fuzz_binary_script(ctx):
-    script = ctx.actions.declare_file(ctx.label.name)
-
-    script_template = """
+_JAZZER_SCRIPT_TEMPLATE = """
 exec "{driver}" \
     --agent_path="{agent}" \
     --cp="{deploy_jar}" \
     --jvm_args="-Djava.library.path={library_path}" \
     "$@"
 """
+
+_JAZZER_OSS_FUZZ_SCRIPT_TEMPLATE = """#!/bin/sh
+# LLVMFuzzerTestOneInput for fuzzer detection.
+this_dir=$(dirname "$0")
+LD_LIBRARY_PATH="$JVM_LD_LIBRARY_PATH":$this_dir \
+ASAN_OPTIONS=$ASAN_OPTIONS:symbolize=1:external_symbolizer_path=$this_dir/llvm-symbolizer:handle_segv=1:allow_user_segv_handler=1:detect_leaks=0 \
+$this_dir/{driver_basename} \
+--agent_path=$this_dir/jazzer_agent_deploy.jar \
+--cp="{deploy_jar}" \
+--jvm_args="-Xmx2048m;-Djava.library.path={library_path}" \
+$@
+"""
+
+def _jazzer_fuzz_binary_script(ctx, is_oss_fuzz):
+    script = ctx.actions.declare_file(ctx.label.name)
 
     # Perform feature detection for
     # https://github.com/bazelbuild/bazel/commit/381a519dfc082d4c62096c4ce77ead1c2e0410d8.
@@ -125,22 +137,34 @@ exec "{driver}" \
         ]
     native_dirs = [path[:path.rfind("/")] for path in native_paths]
 
-    script_content = script_template.format(
-        driver = ctx.executable.driver.short_path,
-        agent = ctx.file._agent.short_path,
-        deploy_jar = ctx.file.target_deploy_jar.short_path,
-        library_path = ":".join(native_dirs),
-    )
+    if is_oss_fuzz:
+        deps_dir = ctx.attr.base_name + "_deps"
+        script_content = _JAZZER_OSS_FUZZ_SCRIPT_TEMPLATE.format(
+            deploy_jar = deps_dir + "/" + ctx.file.target_deploy_jar.basename,
+            driver_basename = "jazzer_driver_asan" if ctx.attr.transitive_native_deps else "jazzer_driver",
+            library_path = deps_dir,
+        )
+    else:
+        script_content = _JAZZER_SCRIPT_TEMPLATE.format(
+            agent = ctx.file._agent.short_path,
+            deploy_jar = ctx.file.target_deploy_jar.short_path,
+            driver = ctx.executable.driver.short_path,
+            library_path = ":".join(native_dirs),
+        )
     ctx.actions.write(script, script_content, is_executable = True)
     return script
 
 def _jazzer_fuzz_binary_impl(ctx):
-    script = _jazzer_fuzz_binary_script(ctx)
+    is_oss_fuzz = ctx.attr._cc_engine.label.name == "oss_fuzz_engine"
+    script = _jazzer_fuzz_binary_script(ctx, is_oss_fuzz)
 
     runfiles = ctx.runfiles()
-    runfiles = runfiles.merge(ctx.attr.driver[DefaultInfo].default_runfiles)
-    runfiles = runfiles.merge(ctx.runfiles([ctx.file._agent]))
-    runfiles = runfiles.merge(ctx.attr.target[0][DefaultInfo].default_runfiles)
+
+    # Driver and agent are provided by OSS-Fuzz, so only include them as
+    # runfiles in normal builds.
+    if not is_oss_fuzz:
+        runfiles = runfiles.merge(ctx.attr.driver[DefaultInfo].default_runfiles)
+        runfiles = runfiles.merge(ctx.runfiles([ctx.file._agent]))
     runfiles = runfiles.merge(ctx.runfiles([ctx.file.target_deploy_jar]))
     for native_dep in ctx.attr.transitive_native_deps:
         runfiles = runfiles.merge(native_dep[DefaultInfo].default_runfiles)
@@ -152,10 +176,21 @@ jazzer_fuzz_binary = rule(
 Rule that creates a binary that invokes Jazzer on the specified target.
 """,
     attrs = {
+        "_cc_engine": attr.label(
+            default = Label("@rules_fuzzing//fuzzing:cc_engine"),
+            doc = "Used to determine whether the binary should be produced " +
+                  "for OSS-Fuzz.",
+        ),
         "_agent": attr.label(
             default = Label("@jazzer//agent:jazzer_agent_deploy.jar"),
             doc = "The Jazzer agent used to instrument the target.",
             allow_single_file = [".jar"],
+        ),
+        "base_name": attr.string(
+            default = "",
+            doc = "The base name of the fuzz test used to form the file names" +
+                  " in the OSS-Fuzz output.",
+            mandatory = True,
         ),
         "driver": attr.label(
             default = Label("@jazzer//driver:jazzer_driver"),
--- a/fuzzing/private/java_utils.bzl
+++ b/fuzzing/private/java_utils.bzl
@@ -138,11 +138,11 @@ def _jazzer_fuzz_binary_script(ctx, is_oss_fuzz):
     native_dirs = [path[:path.rfind("/")] for path in native_paths]
 
     if is_oss_fuzz:
-        deps_dir = ctx.attr.base_name + "_deps"
+        runfiles_dir = ctx.attr.base_name + ".runfiles"
         script_content = _JAZZER_OSS_FUZZ_SCRIPT_TEMPLATE.format(
-            deploy_jar = deps_dir + "/" + ctx.file.target_deploy_jar.basename,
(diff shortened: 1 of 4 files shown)


fmeum:
I rebased the PR onto master.

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -35,6 +57,8 @@ def _oss_fuzz_package_impl(ctx):
             }}
             trap cleanup EXIT
             ln -s "$(pwd)/{binary_path}" "$STAGING_DIR/{base_name}"
+            mkdir "$STAGING_DIR/{binary_deps_dir}"
+            {binary_deps_script}
```
stefanbucur, line 61: Would it be possible to implement the logic of computing these files as part of this script?

Right now the logic of this action is split between shell script fragments and Starlark code, making the code hard to validate upon visual inspection.

It would be great to avoid forming a script out of Starlark-assembled snippets.

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -25,6 +25,28 @@ def _oss_fuzz_package_impl(ctx):
...
     if binary_info.dictionary_file:
         action_inputs.append(binary_info.dictionary_file)
+
+    binary_deps = [
+        runfile
+        for runfile in binary_info.binary_runfiles.files.to_list()
+        if runfile != binary_info.binary_file
+    ]
+    action_inputs += binary_deps
+    binary_deps_dir = ctx.attr.base_name + "_deps"
```
stefanbucur, line 35: I believe the Bazel convention is to have the runfiles of a binary `<name>` as a separate directory `<name>.runfiles` side-by-side with the binary (e.g., as documented [here](https://docs.bazel.build/versions/master/skylark/rules.html#runfiles-location)).

Should we preserve this convention here, too? This way, programs that are aware of their runfiles will be able to automatically discover this directory based on this convention.

stefanbucur (changes requested):
Apologies for the latency!

fuzzing/private/oss_fuzz/package.bzl
fmeum: That makes sense, I will change it. 

We will deviate from what Bazel does in that we currently don't mimic the full runfile structure under `<name>.runfiles`, but only a flat structure where all runfile paths are truncated to their base names. Is that okay or should I extend the logic here?

fuzzing/private/java_utils.bzl
```diff
@@ -90,17 +90,29 @@ def _java_segment_index(path_segments):
...
+_JAZZER_SCRIPT_TEMPLATE = """
 exec "{driver}" \
     --agent_path="{agent}" \
     --cp="{deploy_jar}" \
     --jvm_args="-Djava.library.path={library_path}" \
     "$@"
 """
 
+_JAZZER_OSS_FUZZ_SCRIPT_TEMPLATE = """#!/bin/sh
+# LLVMFuzzerTestOneInput for fuzzer detection.
```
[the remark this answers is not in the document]
fmeum: For OSS-Fuzz to treat a file as a fuzz target, it needs to be executable and contain the ASCII byte string `LLVMFuzzerTestOneInput`, e.g. [here](https://github.com/google/oss-fuzz/blob/30f3a8f1c0f5b072e77d5bea82709db04c53453d/infra/base-images/base-runner/targets_list). Hence all wrapper scripts for languages that are not C/C++, such as Python and Java, use this comment to pass this check.

fuzzing/private/java_utils.bzl
```diff
@@ -90,17 +90,29 @@ def _java_segment_index(path_segments):
...
-    script_template = """
+_JAZZER_SCRIPT_TEMPLATE = """
 exec "{driver}" \
     --agent_path="{agent}" \
     --cp="{deploy_jar}" \
     --jvm_args="-Djava.library.path={library_path}" \
     "$@"
 """
 
+_JAZZER_OSS_FUZZ_SCRIPT_TEMPLATE = """#!/bin/sh
```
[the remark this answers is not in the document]
fmeum: The main differences are the following:

1. The fuzz target should run with the version of Jazzer provided by OSS-Fuzz, not the one it built from source itself.
2. Whereas fuzz targets should probably run with JVM defaults locally, OSS-Fuzz imposes stricter limits on memory usage that need to be followed by the JVM.
3. Java fuzz targets need to deviate from the default `ASAN_OPTIONS` in multiple ways as they use custom SEGV handlers and do not support leak detection.
4. Runfiles support is limited (flat directory instead of full runfile tree), meaning that paths need to be translated. This will always be necessary in some form as relative runfile paths such as `../target/deploy.jar` can never be realized in OSS-Fuzz.
5. ClusterFuzz runs fuzz targets with a different working directory, which requires the `$this_dir` dance.

It would be great if we could find ways to get rid of some of this complexity, but I can't see how we would be able to address all points transparently.

fuzzing/private/java_utils.bzl
```diff
@@ -152,11 +176,22 @@ jazzer_fuzz_binary = rule(
 Rule that creates a binary that invokes Jazzer on the specified target.
 """,
     attrs = {
+        "_cc_engine": attr.label(
+            default = Label("@rules_fuzzing//fuzzing:cc_engine"),
+            doc = "Used to determine whether the binary should be produced " +
+                  "for OSS-Fuzz.",
+        ),
```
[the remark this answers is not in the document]
fmeum: The somewhat subtle difference here is that this attribute is for an engine of a language other than the one of the macro: We specifically need the `cc_engine` here to determine whether we are building for OSS-Fuzz or not. I could see a `cc_engine` being passed as an `engine` attribute to a `jazzer_fuzz_binary` rule being confusing. What do you think?

fuzzing/private/oss_fuzz/package.bzl
fmeum: I'm not good with shell scripts, which is probably why I try to do as much in Starlark as possible. Could you give me some hints how I could do more with the shell here? I guess I could use a bash loop instead of a Starlark one, but how would I go iterate over pairs `(runfile_path, runfile_name)` in Bash?

fmeum:
Thanks for the review, I left a few comments and switched to the `<name>.runfiles` naming convention.

The CI failure is quite unexpected: https://github.com/bazelbuild/rules_fuzzing/pull/146/checks?check_run_id=2581533367
It seems that the LLVM `FuzzedDataProvider` can't find the `<limits>` header.

fuzzing/private/java_utils.bzl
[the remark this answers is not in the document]
stefanbucur: Thanks for clarifying all these. It sounds good to me. Do you think we could define an OSS-Fuzz test project that (a) OSS-Fuzz developers can use as a starting point and (b) can be run as a Github action here through CI Fuzzing? I think we should have an end-to-end test demonstrating that all the pieces work together.

We do have a test project for the C++ rules at https://github.com/google/oss-fuzz/tree/master/projects/bazel-rules-fuzzing-test (CI Fuzzing integration [here](https://github.com/bazelbuild/rules_fuzzing/blob/master/.github/workflows/oss_fuzz.yml)), but I'm not sure the same project can build both C++ and Java fuzz targets at once.

CC @inferno-chromium @jonathanmetzman for their inputs too.

fuzzing/private/java_utils.bzl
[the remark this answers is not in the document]
stefanbucur: Got it. IMO, this is not obvious from this comment alone. We should clearly document gotchas like these, so the next person working on the script does not accidentally remove it.

How about saying "LLVMFuzzerTestOneInput - OSS-Fuzz needs this string literal to appear somewhere in the script so it is recognized as a fuzz target."

fuzzing/private/java_utils.bzl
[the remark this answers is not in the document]
stefanbucur: Oh, I understand now - I did not notice the "cc" part the first time. Maybe state this explicitly in the doc string: "The C++ engine used to determine whether the binary..."

fuzzing/private/java_utils.bzl
```diff
@@ -152,11 +176,22 @@ jazzer_fuzz_binary = rule(
...
         "_agent": attr.label(
             default = Label("@jazzer//agent:jazzer_agent_deploy.jar"),
             doc = "The Jazzer agent used to instrument the target.",
             allow_single_file = [".jar"],
         ),
+        "base_name": attr.string(
+            default = "",
```
stefanbucur, line 190: Nit: If this is mandatory, do we need to set a default here?

stefanbucur (changes requested):

fuzzing/private/oss_fuzz/package.bzl
stefanbucur: > That makes sense, I will change it.
> 
> We will deviate from what Bazel does in that we currently don't mimic the full runfile structure under `<name>.runfiles`, but only a flat structure where all runfile paths are truncated to their base names. Is that okay or should I extend the logic here?

I think that would cause issues, since the runfiles structure is a contract that Bazel provides to the binaries that have data dependencies. Moreover, what happens for distinct paths that share the same base name? I do think we should preserve the actual runfile structure in the OSS-Fuzz packaging.

fuzzing/private/oss_fuzz/package.bzl
stefanbucur: I think you can use the `find` command to iterate over the entire directory tree. You can loop over the command output and run the `ln -s` command to create the symbolic links (you might need to explicitly create the intermediate directories, too, but I'm not 100% sure).

Let me know if you run into any issues - I'm happy to dig further into this.

fuzzing/private/oss_fuzz/package.bzl
fmeum: While it's possible to port the entire runfiles tree to OSS-Fuzz, I think that we would need to go pretty deep here. Basically this requires:
1. Looping over all `short_path`s computing the longest prefix chain of `../`.
2. Moving the actual binary that deep into a synthetic folder structure.
3. Move it's runfiles relative to it based on their `short_path`s, creating directories as needed.
4. Emit a top-level wrapper binary that `cd`s to the binary and executes it.

I can do that and would find it useful, but I'm not sure I'll be able to pull it off in pure bash. Should I start with a mix of bash/Starlark and we clean it up later?

fmeum: This may not be that difficult after all. `short_path` always starts with at most one `../` (in case of a runfiles in another workspace), right?

fmeum:
I will address the comments by splitting off some improved independent parts of this PR and submitting them as individual PRs for easier reviews. #147 and #148 are the first ones.

fmeum:
I have also split out the runfiles tree implementation in #149.

Once the other three PRs have been merged, I will update the current one with the Java-specific part of OSS-Fuzz support, which by then will hopefully require less special casing. All PRs seem to be hit by the same CI issue though.

fmeum:
I will switch this PR to draft state. All its parts that are not Java specific have now been submitted separately as #147 through #151. Once those have been reviewed, I will take care of the (few) Java specific tweaks here.

stefanbucur:
> I will switch this PR to draft state. All its parts that are not Java specific have now been submitted separately as #147 through #151. Once those have been reviewed, I will take care of the (few) Java specific tweaks here.

Sounds good - thanks for seeing through all these changes. Do you have a sense of how much there is left before we can consider this feature complete and be able to cut a release?

fmeum:
What remains is one medium-size PR that should basically only touch `java_utils.bzl`. I don't know yet how much time I'll have this week, but there is not much work left. I would also need to add the OSS-Fuzz test project though before this could be merged.

I only recently learned from Jonathan that the use of options files is discouraged since they supposedly don't apply to CI Fuzz. The first alternative that comes to mind would be a shell script setting environment variables that is called from the wrapper and chosen via a `select` on the `cc_engine`. What do you think?

stefanbucur:
> What remains is one medium-size PR that should basically only touch `java_utils.bzl`. I don't know yet how much time I'll have this week, but there is not much work left. I would also need to add the OSS-Fuzz test project though before this could be merged.

Sounds great!

> I only recently learned from Jonathan that the use of options files is discouraged since they supposedly don't apply to CI Fuzz. The first alternative that comes to mind would be a shell script setting environment variables that is called from the wrapper and chosen via a `select` on the `cc_engine`. What do you think?

Oh that's interesting, I didn't know about options files being deprecated.

Using a shell script might work. I am also curious if Jonathan has any recommended alternative to the options file for environment variables.

fmeum:
The reference: https://github.com/google/oss-fuzz/pull/5910#issuecomment-859541998

fmeum:
I will close this PR in favor of #157 since the respective commits ended up differing dramatically.

PR #147, opened 2021-05-16
Remove unused jazzer_fuzz_binary runfile

fmeum:
This commit removes the following two types of runfiles declared by the
java_binary target passed into the jazzer_fuzz_binary rule which are not
used during the execution of the Jazzer target:

* the local JDK files in @local_jdk (Jazzer launches the JVM directly)
* the individual jars on the runtime classpath (Jazzer executes the
  deploy jar, which contains the contents of all these jars).

### DIFF
--- a/fuzzing/private/java_utils.bzl
+++ b/fuzzing/private/java_utils.bzl
@@ -134,13 +134,35 @@ exec "{driver}" \
     ctx.actions.write(script, script_content, is_executable = True)
     return script
 
+def _is_required_runfile(target, runtime_classpath, runfile):
+    # The jars in the runtime classpath are all merged into the deploy jar and
+    # thus don't need to be included in the runfiles for the fuzzer.
+    if runfile in runtime_classpath:
+        return False
+
+    # A java_binary target has a dependency on the local JDK. Since the Jazzer
+    # driver launches its own JVM, these runfiles are not needed.
+    if runfile.owner != None and runfile.owner.workspace_name == "local_jdk":
+        return False
+    return True
+
+def _filter_target_runfiles(ctx, target):
+    compilation_info = target[JavaInfo].compilation_info
+    runtime_classpath = compilation_info.runtime_classpath.to_list()
+    all_runfiles = target[DefaultInfo].default_runfiles
+    return ctx.runfiles([
+        runfile
+        for runfile in all_runfiles.files.to_list()
+        if _is_required_runfile(target, runtime_classpath, runfile)
+    ])
+
 def _jazzer_fuzz_binary_impl(ctx):
     script = _jazzer_fuzz_binary_script(ctx)
 
     runfiles = ctx.runfiles()
     runfiles = runfiles.merge(ctx.attr.driver[DefaultInfo].default_runfiles)
     runfiles = runfiles.merge(ctx.runfiles([ctx.file._agent]))
-    runfiles = runfiles.merge(ctx.attr.target[0][DefaultInfo].default_runfiles)
+    runfiles = runfiles.merge(_filter_target_runfiles(ctx, ctx.attr.target[0]))
     runfiles = runfiles.merge(ctx.runfiles([ctx.file.target_deploy_jar]))
     for native_dep in ctx.attr.transitive_native_deps:
         runfiles = runfiles.merge(native_dep[DefaultInfo].default_runfiles)


fmeum:
The build failure is the same weird one we had seen before on #146. @stefanbucur Do you have an idea what could have broken here?

stefanbucur (commented):
The change looks ok, but wanted to double-check before approving: If the java_fuzz_test() itself has data dependencies, will they still be added to the target runfiles if we remove this statement?

fmeum:
> The change looks ok, but wanted to double-check before approving: If the java_fuzz_test() itself has data dependencies, will they still be added to the target runfiles if we remove this statement?

Probably not, but I think `data` dependencies were not even supported previous to this change when running the fuzz test locally within Bazel due to it only using the `_deploy.jar` target of the `java_binary`. The dependency on the `java_binary` is only there to help collect transitive native library dependencies automatically with an upcoming stable release of Bazel.

Data dependencies on `java_binary` targets are, at least in my experience, used rarely (if ever) as there is the Java specific [`resources`](https://docs.bazel.build/versions/master/be/java.html#java_binary.resources) attribute that is supported natively by Java and embeds the data directly into the `_deploy.jar`. For runfiles, jar targets would need to depend on the Bazel-specific https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java to find them.

I could see a case for data dependencies if the `jar` needs to run native binaries (rather than using native libraries directly). I would say it's quite unlikely that this use case ever shows up for fuzz tests, but if you would prefer support for them I could think about how to add it.

fmeum:
I checked and Bazel does not add the `data` dependencies of the `java_binary` to the `_deploy.jar` target. I'm pretty sure that's on purpose: In the Java world, one would rather use `resources` instead of `data` as that 1) is supported natively by Java and 2) ends up in the `_deploy.jar`.

The only use case for `data` on a `java_binary` I could see is a native binary that can't be used as a library. But I can hardly see that playing a role for fuzz tests.

What do you think? I could add a descriptive failure message if the `data` attribute of `java_fuzz_test` is set to a non-`None` value.

stefanbucur:
> I checked and Bazel does not add the `data` dependencies of the `java_binary` to the `_deploy.jar` target. I'm pretty sure that's on purpose: In the Java world, one would rather use `resources` instead of `data` as that 1) is supported natively by Java and 2) ends up in the `_deploy.jar`.
> 
> The only use case for `data` on a `java_binary` I could see is a native binary that can't be used as a library. But I can hardly see that playing a role for fuzz tests.
> 
> What do you think? I could add a descriptive failure message if the `data` attribute of `java_fuzz_test` is set to a non-`None` value.

Sorry for asking more questions, I'm not super familiar with the Java rules and wanted to make sure I understand the issue.

So where does the `_deploy.jar` rule come into play here? If I do `bazel run :JavaFuzzer_run`, will it execute the `deploy.jar` target? If it runs the plain Java binary, I would expect the data dependencies to be available in the target runfiles.

fmeum:
> Sorry for asking more questions, I'm not super familiar with the Java rules and wanted to make sure I understand the issue.
> 
> So where does the `_deploy.jar` rule come into play here? If I do `bazel run :JavaFuzzer_run`, will it execute the `deploy.jar` target? If it runs the plain Java binary, I would expect the data dependencies to be available in the target runfiles.

If you have a `java_binary` rule `foobar`, then `foobar` can have `data` dependencies and these will be available to the target when run with `bazel run //:foobar`. They could then be accessed via the helper library [Runfiles.java](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java). However, Java already has the mechanism of resource files, which basically serve the same purpose and are natively supported by the language.

Jazzer, the Java fuzzer, needs to start its own JVM in order to monitor and steer the execution of this fuzz target. This is why it consists of a driver binary that receives the fuzz target as a data dependency. But this also means that it requires the fuzz target to be "bundled up" in a way that it can be launched easily using the [Java Invocation API](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/invocation.html). That is where the Bazel supplied `foobar_deploy.jar` target comes into play, which achieves exactly that requiring only a single classpath entry, the path to the .jar.
In short, `bazel run //:JavaFuzzer_run` will run the `jazzer_driver`, which in turn spins up a JVM with classpath set to the `JavaFuzzer_raw__deploy.jar` target. The actual Bazel magic involved in the `JavaFuzzer_raw_` target and its runfiles tree is never invoked.

The fact that the `<name>_deploy.jar` targets are specifically meant for deployment but don't include `data` dependencies tells me that Bazel devs want users to use `resources` instead. While `data` may be useful for `java_binary` targets that are supposed to be executed during the build, they would seem quite out-of-place on something build for the `target` cfg and especially so for a fuzz test.

fmeum:
I rebased the PR onto the CI fix and added a descriptive failure message when someone tries to use the `data` attribute.

fmeum:
I forgot to add the native libraries to the runfiles in the Bazel 5 case and have marked this as draft until that is fixed.

fuzzing/private/fuzz_test.bzl
```diff
@@ -230,6 +231,10 @@ def java_fuzz_test(
           binary rule.
     """
 
+    if data != None:
+        fail("java_fuzz_test does not support the 'data' attribute, use " +
+             "'resources' instead")
```
stefanbucur, line 236: Nit: Instead of defining a new attribute to the signature of the macro, consider calling `binary_kwargs.pop()` and checking the returned value.

stefanbucur:
> I forgot to add the native libraries to the runfiles in the Bazel 5 case and have marked this as draft until that is fixed.

Sounds good! Let me know when I can take another look.

stefanbucur:
> > Sorry for asking more questions, I'm not super familiar with the Java rules and wanted to make sure I understand the issue.
> > So where does the `_deploy.jar` rule come into play here? If I do `bazel run :JavaFuzzer_run`, will it execute the `deploy.jar` target? If it runs the plain Java binary, I would expect the data dependencies to be available in the target runfiles.
> 
> If you have a `java_binary` rule `foobar`, then `foobar` can have `data` dependencies and these will be available to the target when run with `bazel run //:foobar`. They could then be accessed via the helper library [Runfiles.java](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java). However, Java already has the mechanism of resource files, which basically serve the same purpose and are natively supported by the language.
> 
> Jazzer, the Java fuzzer, needs to start its own JVM in order to monitor and steer the execution of this fuzz target. This is why it consists of a driver binary that receives the fuzz target as a data dependency. But this also means that it requires the fuzz target to be "bundled up" in a way that it can be launched easily using the [Java Invocation API](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/invocation.html). That is where the Bazel supplied `foobar_deploy.jar` target comes into play, which achieves exactly that requiring only a single classpath entry, the path to the .jar.
> In short, `bazel run //:JavaFuzzer_run` will run the `jazzer_driver`, which in turn spins up a JVM with classpath set to the `JavaFuzzer_raw__deploy.jar` target. The actual Bazel magic involved in the `JavaFuzzer_raw_` target and its runfiles tree is never invoked.
> 
> The fact that the `<name>_deploy.jar` targets are specifically meant for deployment but don't include `data` dependencies tells me that Bazel devs want users to use `resources` instead. While `data` may be useful for `java_binary` targets that are supposed to be executed during the build, they would seem quite out-of-place on something build for the `target` cfg and especially so for a fuzz test.

Thanks a lot for taking the time to write this detailed explanation. I found it very helpful.

fuzzing/private/fuzz_test.bzl
fmeum: That's good advice, thanks!

fmeum:
I force pushed a completely new approach which selectively filters known unused runfiles from the target, but preserves all others. This is mostly needed so that the OSS-Fuzz archives don't each contain a complete copy of the local JDK.

stefanbucur (approved):

Issue #104, opened 2021-01-06
Propagate fuzzing options to the OSS-Fuzz package, through the creation of the options file.


stefanbucur:

fmeum:
Is there more that needs to be done here now that #148 has been merged? `.options` files are not exposed via the `*_fuzz_test` macros yet, but this also doesn't seem to be needed for C++ fuzz tests since `max_len` should rather be implemented in the fuzz test directly.

stefanbucur:
You're right, I think this is now addressed. Thanks!

PR #148, opened 2021-05-16
Add support for .options files to fuzzing_binary

fmeum:
.options files can be used to supply custom options to sanitizers. While
they are rarely used for C++ fuzz targets (and thus not exposed via the
cc_fuzz_test macro for now), they are necessary to run JVM fuzz targets
correctly and will be used internally by the java_fuzz_test macro in a
follow-up commit.

This generalizes parts of #146.


fmeum:
If we expose this functionality through the `{cc, java}_fuzz_test` macro, this could resolve #104. @stefanbucur What do you think?

stefanbucur:
> If we expose this functionality through the `{cc, java}_fuzz_test` macro, this could resolve #104. @stefanbucur What do you think?

I would like to understand a bit what design alternatives we have here. If I remember correctly, one idea behind #104 was to automatically generate the .options file for OSS-Fuzz from cc_fuzz_test() attributes. This way, developers don't need to maintain one more file with its own file format.

Can we do the same here? What does the developer need to provide in an .options file for Java fuzz targets?

fmeum:
For Java targets the developer will most likely never need to supply their own `.options` file. Rather, the file would collect all or at least most of the `ASAN_OPTIONS` that were previously added to the shell wrapper conditionally [here](https://github.com/bazelbuild/rules_fuzzing/pull/146/files#diff-ce832be004e8f7974f5a41d2851ba12b4b44facc021cdf5ee9c7625becab6c1eR105).

I believe the official guidance on length restrictions for C++ fuzz tests is to prefix them with`if (Size > 4096) return;` style checks rather than use the libFuzzer-dependent `max_len`, which is otherwise the most obvious candidate for options to me. Do you have some particular options for `cc_fuzz_test` targets in mind? If so, we could start with exposing just those (and not exposing any options for `java_fuzz_test` for now)?

stefanbucur:
Meta-comment: This week I've been away and my replies are slower. Sorry for falling behind with code reviews.

stefanbucur:
> For Java targets the developer will most likely never need to supply their own `.options` file. Rather, the file would collect all or at least most of the `ASAN_OPTIONS` that were previously added to the shell wrapper conditionally [here](https://github.com/bazelbuild/rules_fuzzing/pull/146/files#diff-ce832be004e8f7974f5a41d2851ba12b4b44facc021cdf5ee9c7625becab6c1eR105).

Ah, I see. How will the ASAN options be passed in that file? I've only seen so far libFuzzer flags, but not ASAN environment options.

> 
> I believe the official guidance on length restrictions for C++ fuzz tests is to prefix them with`if (Size > 4096) return;` style checks rather than use the libFuzzer-dependent `max_len`, which is otherwise the most obvious candidate for options to me. Do you have some particular options for `cc_fuzz_test` targets in mind? If so, we could start with exposing just those (and not exposing any options for `java_fuzz_test` for now)?

Yeah, I can't think of particular C++ options needed at this point (and I agree with you on the `max_len` use case).

fmeum:
> Ah, I see. How will the ASAN options be passed in that file? I've only seen so far libFuzzer flags, but not ASAN environment options.

The `java-example` project currently uses a mix of ASan options passed [via the environment variable](https://github.com/CodeIntelligenceTesting/oss-fuzz/blob/a9fe789db7b3535869285fbd78f329d81612eb30/projects/java-example/build.sh#L45) and a [.options file](https://github.com/CodeIntelligenceTesting/oss-fuzz/blob/master/projects/java-example/default.options) (purely for historical reasons). These are in fact interchangeable (except that the .options file does not support bash variable expansion): https://github.com/CodeIntelligenceTesting/oss-fuzz/blob/a9fe789db7b3535869285fbd78f329d81612eb30/infra/base-images/base-runner/run_fuzzer#L80.

stefanbucur:
> Meta-comment: This week I've been away and my replies are slower. Sorry for falling behind with code reviews.

Thanks a lot for your patience with these reviews - I'm now back and will go through these as soon as possible.

fuzzing/private/binary.bzl
```diff
@@ -36,6 +36,7 @@ Provider for storing information about a fuzz test binary.
         "corpus_dir": "The directory of the corpus files used as input seeds.",
         "dictionary_file": "The dictionary file to use in fuzzing runs.",
         "engine_info": "The `FuzzingEngineInfo` provider of the fuzzing engine used in the fuzz test.",
+        "options_file": "The .options file to use in OSS-Fuzz.",
```
stefanbucur, line 39: I think attaching an options file to the FuzzingBinary makes sense, but I would like to decouple this concept from OSS-Fuzz as much as possible.

In this case, let's not define this file in terms of how it is being used (which is subject to change in the future).

I would suggest the following description: "A file containing fuzzing engine and sanitizer options to use during execution. The file follows the INI format defined in OSS-Fuzz." Ideally, we'd also include a link to the file format documentation - do we have one?

fuzzing/private/binary.bzl
```diff
@@ -158,6 +162,10 @@ The instrumentation is controlled by the following flags:
             doc = "A dictionary file to use in fuzzing runs.",
             allow_single_file = True,
         ),
+        "options": attr.label(
+            doc = "A file with INI-style options for OSS-Fuzz.",
```
stefanbucur, line 166: Same observation here: Describe first what the file contains, independently of how it is being used. You can have a clarifying sentence at the end, e.g., "Currently, this file is only used in OSS-Fuzz runs."

fuzzing/private/binary.bzl
```diff
@@ -198,6 +206,10 @@ be incorporated in the target configuration (e.g., on the command line or the
...
+            allow_single_file = True,
+        ),
```
stefanbucur, line 212: I don't think this duplication scales. Should we define the shared attributes in a separate dict variable that we merge into each rule definition?

stefanbucur (changes requested):

fuzzing/private/binary.bzl
fmeum: Unfortunately there is no such documentation. `.options` files are only mentioned in passing when the docs talk about shared dictionary files. I like your explanation though and am using it now.

fuzzing/private/binary.bzl
fmeum: I did using skylib's `dicts.add`.

fmeum:
I pushed a new commit to address your comments.

fuzzing/private/binary.bzl
```diff
@@ -36,6 +37,7 @@ Provider for storing information about a fuzz test binary.
         "corpus_dir": "The directory of the corpus files used as input seeds.",
         "dictionary_file": "The dictionary file to use in fuzzing runs.",
         "engine_info": "The `FuzzingEngineInfo` provider of the fuzzing engine used in the fuzz test.",
+        "options_file": "A file containing fuzzing engine and sanitizer options to use during execution. The file loosely follows the INI format and currently only applies to OSS-Fuzz.",
```
fmeum, line 40: The line above exceeds the usual line length limit, so I kept this as one line as well. Should I wrap both instead?

stefanbucur: Sure, wrapping both makes sense to me.

stefanbucur (approved):

fuzzing/private/binary.bzl
fmeum: Done.

PR #149, opened 2021-05-17
Add full runfiles tree support

### LINKED ISSUE #100: Support packaging runfiles in the OSS-Fuzz packaging rule.

fmeum:
The runfiles tree of a fuzzing_binary is now packaged correctly for
OSS-Fuzz.

For a fuzz test foo in workspace bar, the runfiles are resolved relative
to the foo.runfiles/bar directory in the output .tar. A wrapper script
ensures that the real fuzz test binary is called with the correct
working directory.

The workspace_name of the binary label attribute in fuzzing_binary is
needed in the construction of the binary's runfiles tree and thus added
to the FuzzingBinaryInfo provider.

Closes #100.

### DIFF
--- a/examples/runfiles_fuzz_test.cc
+++ b/examples/runfiles_fuzz_test.cc
@@ -18,24 +18,27 @@
 #include <cstdint>
 
 #include <fstream>
-#include <memory>
+#include <iostream>
 #include <string>
 
 #include "tools/cpp/runfiles/runfiles.h"
 
-using bazel::tools::cpp::runfiles::Runfiles;
+using ::bazel::tools::cpp::runfiles::Runfiles;
 
 namespace {
-  std::unique_ptr<Runfiles> runfiles = nullptr;
+  Runfiles *runfiles = nullptr;
 }
 
 extern "C" void LLVMFuzzerInitialize(int *argc, char ***argv) {
   std::string error;
-  runfiles = std::unique_ptr<Runfiles>(Runfiles::Create((*argv)[0], &error));
+  runfiles = Runfiles::Create((*argv)[0], &error);
+  if (runfiles == nullptr) {
+    std::cerr << error;
+    abort();
+  }
 }
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-  if (runfiles == nullptr) abort();
   std::string path = runfiles->Rlocation("rules_fuzzing/examples/corpus_0.txt");
   if (path.empty()) abort();
   std::ifstream in(path);
--- a/examples/runfiles_fuzz_test.cc
+++ b/examples/runfiles_fuzz_test.cc
@@ -1,0 +1,44 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// A fuzz target that exits if it doesn't find a declared runfile.
+
+#include <cstddef>
+#include <cstdint>
+
+#include <fstream>
+#include <memory>
+#include <string>
+
+#include "tools/cpp/runfiles/runfiles.h"
+
+using bazel::tools::cpp::runfiles::Runfiles;
+
+namespace {
+  std::unique_ptr<Runfiles> runfiles = nullptr;
+}
+
+extern "C" void LLVMFuzzerInitialize(int *argc, char ***argv) {
+  std::string error;
+  runfiles = std::unique_ptr<Runfiles>(Runfiles::Create((*argv)[0], &error));
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  if (runfiles == nullptr) abort();
+  std::string path = runfiles->Rlocation("rules_fuzzing/examples/corpus_0.txt");
+  if (path.empty()) abort();
+  std::ifstream in(path);
+  if (!in.good()) abort();
+  return 0;
+}
--- a/fuzzing/private/oss_fuzz/package.bzl
+++ b/fuzzing/private/oss_fuzz/package.bzl
@@ -15,26 +15,47 @@
 """Rule for packaging fuzz tests in the expected OSS-Fuzz format."""
 
 load("//fuzzing/private:binary.bzl", "FuzzingBinaryInfo")
+load("//fuzzing/private:util.bzl", "runfile_path")
 
 def _oss_fuzz_package_impl(ctx):
     output_archive = ctx.actions.declare_file(ctx.label.name + ".tar")
     binary_info = ctx.attr.binary[FuzzingBinaryInfo]
 
-    action_inputs = [binary_info.binary_file]
+    binary_runfiles = binary_info.binary_runfiles.files.to_list()
+    archive_inputs = binary_runfiles
+
+    runfiles_manifest = ctx.actions.declare_file(ctx.label.name + "_runfiles")
+    runfiles_manifest_content = "".join([
+        "{runfile_path} {real_path}\n".format(
+            real_path = runfile.path,
+            runfile_path = runfile_path(ctx, runfile),
+        )
+        for runfile in binary_runfiles
+        if runfile != binary_info.binary_file
+    ])
+    ctx.actions.write(runfiles_manifest, runfiles_manifest_content, False)
+    archive_inputs.append(runfiles_manifest)
+
     if binary_info.corpus_dir:
-        action_inputs.append(binary_info.corpus_dir)
+        archive_inputs.append(binary_info.corpus_dir)
     if binary_info.dictionary_file:
-        action_inputs.append(binary_info.dictionary_file)
+        archive_inputs.append(binary_info.dictionary_file)
     ctx.actions.run_shell(
         outputs = [output_archive],
-        inputs = action_inputs,
+        inputs = archive_inputs,
         command = """
+            set -e
             declare -r STAGING_DIR="$(mktemp --directory -t oss-fuzz-pkg.XXXXXXXXXX)"
             function cleanup() {{
                 rm -rf "$STAGING_DIR"
             }}
             trap cleanup EXIT
             ln -s "$(pwd)/{binary_path}" "$STAGING_DIR/{base_name}"
+            while IFS= read -r line; do
+              IFS=' ' read -r link target <<< "$line"
+              mkdir -p "$(dirname "$STAGING_DIR/{binary_runfiles_dir}/$link")"
+              ln -s "$(pwd)/$target" "$STAGING_DIR/{binary_runfiles_dir}/$link"
+            done <{runfiles_manifest_path}
             if [[ -n "{corpus_dir}" ]]; then
                 pushd "{corpus_dir}" >/dev/null
                 zip --quiet -r "$STAGING_DIR/{base_name}_seed_corpus.zip" ./*
@@ -50,10 +71,12 @@ def _oss_fuzz_package_impl(ctx):
         """.format(
             base_name = ctx.attr.base_name,
             binary_path = binary_info.binary_file.path,
+            binary_runfiles_dir = ctx.attr.base_name + ".runfiles",
             corpus_dir = binary_info.corpus_dir.path if binary_info.corpus_dir else "",
             dictionary_path = binary_info.dictionary_file.path if binary_info.dictionary_file else "",
             options_path = binary_info.options_file.path if binary_info.options_file else "",
             output = output_archive.path,
+            runfiles_manifest_path = runfiles_manifest.path,
         ),
     )
     return [DefaultInfo(files = depset([output_archive]))]
@@ -62,9 +85,6 @@ oss_fuzz_package = rule(
     implementation = _oss_fuzz_package_impl,
     doc = """
 Packages a fuzz test in a TAR archive compatible with the OSS-Fuzz format.
-
-> NOTE: The current implementation does not yet support packaging the
-> binary runfiles.
 """,
     attrs = {
         "binary": attr.label(
--- a/fuzzing/private/oss_fuzz/package.bzl
+++ b/fuzzing/private/oss_fuzz/package.bzl
@@ -68,9 +68,16 @@ def _oss_fuzz_package_impl(ctx):
                 ln -s "$(pwd)/{options_path}" "$STAGING_DIR/{base_name}.options"
             fi
             tar -chf "{output}" -C "$STAGING_DIR" .
+            # Add a relative symlink to the fuzz test binary to its runfiles.
+            declare -r BINARY_RUNFILES_PATH="$STAGING_DIR/{binary_runfiles_dir}/{binary_runfile_path}"
+            declare -r BINARY_RELATIVE_PATH="$(realpath -m -s --relative-to="$(dirname $BINARY_RUNFILES_PATH)" "$STAGING_DIR/{base_name}")"
+            mkdir -p "$(dirname "$BINARY_RUNFILES_PATH")"
+            ln -s "$BINARY_RELATIVE_PATH" "$BINARY_RUNFILES_PATH"
+            tar -rf "{output}" -C "$STAGING_DIR" "./{binary_runfiles_dir}/{binary_runfile_path}"
         """.format(
             base_name = ctx.attr.base_name,
             binary_path = binary_info.binary_file.path,
+            binary_runfile_path = runfile_path(ctx, binary_info.binary_file),
             binary_runfiles_dir = ctx.attr.base_name + ".runfiles",
             corpus_dir = binary_info.corpus_dir.path if binary_info.corpus_dir else "",
             dictionary_path = binary_info.dictionary_file.path if binary_info.dictionary_file else "",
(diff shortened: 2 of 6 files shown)


fmeum:
I personally prefer the current mix of Starlark to Bash to doing strictly more in Bash, simply because it's harder to make mistakes when using the mechanisms intended by Bazel to deal with runfiles. 

If we use `find` to discover runfiles, wouldn't that open us up to inconsistencies between the files in the runfiles tree and the information in the `FuzzingBinaryInfo` provider's `binary_runfiles`? Not saying there will be any, but we would no longer have a single source of truth.

fmeum:
Note: The CI Fuzz check currently fails as supporting runfile trees for C++ targets requires an OSS-Fuzz change I'll yet have to submit.

fmeum:
I rebased this onto the CI fix, remove unnecessary escaping from the emitted wrapper script and added support for runfiles libraries such as [Runfiles.java](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java).

fuzzing/private/binary.bzl
```diff
@@ -36,6 +36,7 @@ Provider for storing information about a fuzz test binary.
         "corpus_dir": "The directory of the corpus files used as input seeds.",
         "dictionary_file": "The dictionary file to use in fuzzing runs.",
         "engine_info": "The `FuzzingEngineInfo` provider of the fuzzing engine used in the fuzz test.",
+        "workspace_name": "The name of the workspace of the fuzz test executable.",
```
stefanbucur, line 39: I'm curious why this field is actually needed. According to the [`File.short_path` documentation](https://docs.bazel.build/versions/master/skylark/lib/File.html#short_path), `short_path` should be sufficient to determine the runfile location.

stefanbucur, line 39: Assuming this field is actually needed (see my other comment), I think this name is non-specific and inconsistent with the other fields. How about calling it `binary_workspace` and placing it along the other `binary_*` fields in this structure?

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -20,7 +20,24 @@ def _oss_fuzz_package_impl(ctx):
     output_archive = ctx.actions.declare_file(ctx.label.name + ".tar")
     binary_info = ctx.attr.binary[FuzzingBinaryInfo]
 
-    action_inputs = [binary_info.binary_file]
+    action_inputs = binary_info.binary_runfiles.files.to_list()
```
stefanbucur, line 23: Why is this variable called "action_inputs"?

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -29,12 +46,28 @@ def _oss_fuzz_package_impl(ctx):
...
+            # different from the containing directory.
+            echo "this_dir=\\$(dirname "\\$0")
+# Allow the Bazel runfiles libraries to find the tree.
+export RUNFILES_DIR=\\$(readlink -f "\\$this_dir/{base_name}.runfiles")
+# LLVMFuzzerTestOneInput - OSS-Fuzz needs this string literal
+# to appear somewhere in the script so it is recognized as a
+# fuzz target.
+cd \\$this_dir/{binary_workdir}
+exec {binary_short_path} \\$@" \
+                > "$STAGING_DIR/{base_name}"
```
stefanbucur, line 69: I think there are two opportunities to simplify this implementation:

1. Instead of creating a shell script of concatenated fragments, you can write the runfiles file mapping to a separate file (we can call it a runfiles manifest), containing pairs of paths on each line. You can use a separate file writing action for this to an intermediate file. Inside the main shell script, you can have a simple while loop that iterates and `read`s each line into a pair of variables and executes the symbolic link creation.
2. Similarly, you can generate the Bash launcher as a separate file using another Bazel action. This way, you can avoid the current Bash-within-Bash situation, which can be confusing for a reader.

WDYT?

stefanbucur (changes requested):

fuzzing/private/binary.bzl
fmeum: While `short_path` is sufficient to *use* the runfiles tree, the information contained in it does not suffice to *create* it: Since the `short_path`s aren't relative to the runfiles tree root, but the subdirectory corresponding to the workspace of the target, we cannot learn the name of the workspace from any `short_path`. We don't really need the workspace name as any other unique name (distinct from all external workspaces so that there are no collisions) would do, but deterministically deriving such a name just seems more difficult than passing through the real one.

fuzzing/private/oss_fuzz/package.bzl
fmeum: It was called that before I touched the code and is appended to below, but I am (ab)using its intermediate value to deal with `binary_runfiles` here. I will introduce a new variable that holds the intermediate result only and is not mutated.

fuzzing/private/oss_fuzz/package.bzl
fmeum: That sounds good to me. Could you tell me how to write a Bash while loop that moves each line into a pair of variables without any gotchas (spaces in filenames and such things)? I'm not sure I would get this right myself.

stefanbucur: > That sounds good to me. Could you tell me how to write a Bash while loop that moves each line into a pair of variables without any gotchas (spaces in filenames and such things)? I'm not sure I would get this right myself.

Hmm... very good point about spaces and other special characters :/ My original thought was that we could just have lines of the form `${target} ${link}` and we could use a construct of the form `while read -r target link; do ...; done <{runfiles_manifest}`. But it won't work if there are spaces in the file paths.

However, the ":" character appears to be [invalid in Bazel names](https://docs.bazel.build/versions/master/build-ref.html#name), so perhaps we can use that as a separator? I.e., generate a file where each line is of form `${target}:${link}`, then the shell loop can look like:

```
while IFS=':' read -r target link; do
  # Use here the $target and $link variables.
  mkdir -p ...
  ln -s ...
done <{runfiles_manifest}
```

fuzzing/private/binary.bzl
```diff
@@ -121,6 +123,9 @@ def _fuzzing_binary_impl(ctx):
             corpus_dir = ctx.file.corpus,
             dictionary_file = ctx.file.dictionary,
             engine_info = ctx.attr.engine[FuzzingEngineInfo],
+            # If used within rules_fuzzing, the workspace_name of the binary can
+            # be the empty string.
```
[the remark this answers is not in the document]
fmeum: It would be, as we need to create a distinct `binary_workdir`. But this situation should not happen: If used from another repository with not workspace name, `binary.label.workspace_name` will be `__main__` based on my experiments. The only case where it is the empty string is when used from `rules_fuzzing` itself, which we work around by substituting `ctx.workspace_name` in that case.

fuzzing/private/oss_fuzz/package.bzl
fmeum: I pushed new commits, the second of which implements the approach using a manifest file and bash while loop. I decided to use the format used by the Bazel runfiles manifest itself. This requires multiple values for `IFS`, but should be relatively gotcha proof.

While I personally slightly prefer the approach of the first commit, I am fine with both.

fmeum:
@stefanbucur The OSS-Fuzz CI Fuzz Test reveals a somewhat fundamental problem: The OSS-Fuzz bad build checks for C/C++ run on the wrapper rather than the real fuzz test binary now and thus fail. Do you have an idea how we can preserve these checks but still offer runfiles support also for C/C++?

stefanbucur:
> @stefanbucur The OSS-Fuzz CI Fuzz Test reveals a somewhat fundamental problem: The OSS-Fuzz bad build checks for C/C++ run on the wrapper rather than the real fuzz test binary now and thus fail. Do you have an idea how we can preserve these checks but still offer runfiles support also for C/C++?
> 
> Edit: A straightforward way to fix this issue would be to include a meta comment in the wrapper (e.g. `# REAL_BINARY: foo.runfiles/foo/bar`) that points the check build script to the real binary to check. What do you think?

Hmm... I had not noticed that now both C++ and Java binaries get wrapper scripts in the OSS-Fuzz package.

I believe the C++ targets actually don't need the wrapper scripts: I think there is an accepted convention that a Bazel binary `<binary>` can find its runfiles in a `<binary>.runfiles` directory side-by-side.

Is there a way we could make the same assumption about the Java executables packaged in OSS-Fuzz? This way, we could get rid entirely of the wrapper script for all languages.

fmeum:
> I believe the C++ targets actually don't need the wrapper scripts: I think there is an accepted convention that a Bazel binary `<binary>` can find its runfiles in a `<binary>.runfiles` directory side-by-side.

Are you sure about this? The official runfiles libraries (e.g. [the one for Java](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java)) seem to make no such assumption and rather take the runfiles dir from an environment variable which we would need to set. Furthermore, ClusterFuzz currently doesn't launch the fuzz tests with the expected working directory, so finding a sibling directory would require some special effort in the binary.

Assuming that we can fix the problem with the bad build check, do you see other problems with the wrappers for C++ targets?

stefanbucur:
> > I believe the C++ targets actually don't need the wrapper scripts: I think there is an accepted convention that a Bazel binary `<binary>` can find its runfiles in a `<binary>.runfiles` directory side-by-side.
> 
> Are you sure about this? 

Well, I do remember seeing some C++ code that was basically searching for the `.runfiles` dir in a few standard places, but I can't seem to find it at this point, so now I'm no longer sure :/

> The official runfiles libraries (e.g. [the one for Java](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java)) seem to make no such assumption and rather take the runfiles dir from an environment variable which we would need to set. 

Oh, that is interesting, thanks for the pointer!

> Furthermore, ClusterFuzz currently doesn't launch the fuzz tests with the expected working directory, so finding a sibling directory would require some special effort in the binary.

Is the current working directory being set a requirement for the Java targets?

If that one is not required, I'm wondering if it would not be simpler for OSS-Fuzz itself to simply always set the `RUNFILES_DIR` variable to `<binary_name>.runfiles`. If the fuzz test is a Bazel target with runfiles, it will use this variable to find runfiles, otherwise it will just be ignored.

This way, we can avoid this level of indirection in our fuzz test packaging.

> 
> Assuming that we can fix the problem with the bad build check, do you see other problems with the wrappers for C++ targets?

I personally don't, but I also wasn't sure the shell script would violate any other implicit expectations in OSS-Fuzz or Clusterfuzz (the environment that executes the fuzz tests).

So I'm not strongly opposed to the idea, but if there is a (clean) way to avoid that, I'd give that a chance first.

fmeum:
@stefanbucur We were both right: Top-level binaries (those run with `bazel run`) look for a sibling `.runfiles` directory, all binaries launched by those expect environment variables to be set. The working directory doesn't matter. I will rework this PR to follow this convention, which will mean no changes at all to the top-level fuzz test binary at all.

The Jazzer fuzz test binary will have to include the Bazel runfiles bash library, but that will be part of a separate PR.

stefanbucur:
> @stefanbucur We were both right: Top-level binaries (those run with `bazel run`) look for a sibling `.runfiles` directory, all binaries launched by those expect environment variables to be set. The working directory doesn't matter. I will rework this PR to follow this convention, which will mean no changes at all to the top-level fuzz test binary.

That would be awesome - thanks a lot for digging deeper into this issue.

> 
> The Jazzer fuzz test binary will have to include the Bazel runfiles bash library, but that will be part of a separate PR.

Sounds good to me!

fmeum:
@stefanbucur As the top-level binary also needs to be included in the runfiles tree, we have a choice between duplicating it and creating the top-level binary as a symlink into the runfiles tree. What do you prefer?

stefanbucur:
> @stefanbucur As the top-level binary also needs to be included in the runfiles tree

I know the binary gets symlinked by Bazel in the runfiles tree, but I was not aware it is actually needed at runtime, especially if the executable is invoked from the top-level location. In other (C++) projects, I remember being able to package the entire runfiles tree minus the binary itself and did not run into issues.

Do you know if the binary is required in our case, too?

> we have a choice between duplicating it and creating the top-level binary as a symlink into the runfiles tree. What do you prefer?

If we cannot avoid including the binary, then having a relative symlink makes sense to me. I am not sure though how that would interfere with the tarball creation - would the archiver be able to package the symlink instead of the file it points to?

fmeum:
> > @stefanbucur As the top-level binary also needs to be included in the runfiles tree
> 
> I know the binary gets symlinked by Bazel in the runfiles tree, but I was not aware it is actually needed at runtime, especially if the executable is invoked from the top-level location. In other (C++) projects, I remember being able to package the entire runfiles tree minus the binary itself and did not run into issues.
> 
> Do you know if the binary is required in our case, too?

It (let's call it A) is needed if it accesses a runfile binary B at runtime, which again wants to access A. This is possible in Bazel, but of course extremely unlikely in our scenario. But I don't think it's worth deviating from the real Bazel runfile directory layout just to do with one file less. What do you think?

> 
> > we have a choice between duplicating it and creating the top-level binary as a symlink into the runfiles tree. What do you prefer?
> 
> If we cannot avoid including the binary, then having a relative symlink makes sense to me. I am not sure though how that would interfere with the tarball creation - would the archiver be able to package the symlink instead of the file it points to?

Yes, it will preserve the symlink and unpack it correctly (on Unix, Windows is a different story).

stefanbucur:
> > > @stefanbucur As the top-level binary also needs to be included in the runfiles tree
> > 
> > 
> > I know the binary gets symlinked by Bazel in the runfiles tree, but I was not aware it is actually needed at runtime, especially if the executable is invoked from the top-level location. In other (C++) projects, I remember being able to package the entire runfiles tree minus the binary itself and did not run into issues.
> > Do you know if the binary is required in our case, too?
> 
> It (let's call it A) is needed if it accesses a runfile binary B at runtime, which again wants to access A. This is possible in Bazel, but of course extremely unlikely in our scenario. But I don't think it's worth deviating from the real Bazel runfile directory layout just to do with one file less. What do you think?

Sounds good to me then. We can create the top-level binary as a symlink into the runfiles tree. Thanks again for investigating this so thoroughly.

> 
> > > we have a choice between duplicating it and creating the top-level binary as a symlink into the runfiles tree. What do you prefer?
> > 
> > 
> > If we cannot avoid including the binary, then having a relative symlink makes sense to me. I am not sure though how that would interfere with the tarball creation - would the archiver be able to package the symlink instead of the file it points to?
> 
> Yes, it will preserve the symlink and unpack it correctly (on Unix, Windows is a different story).

Great - Given that the OSS-Fuzz environment is Unix-based, this is good enough for now.

fmeum:
I pushed new commits that work on OSS-Fuzz without changes. Making the top-level fuzz test binary a symlink does not currently work in OSS-Fuzz and would require touching at least three different parts of the code. I'm thus putting the real binary in the top-level directory rather than the runfiles directory. The second commit adds a symlink to it in the runfiles tree. I marked it as optional since you need to decide whether the complete runfiles structure is worth the added complexity.

examples/runfiles_fuzz_test.cc
```diff
@@ -0,0 +1,44 @@
...
+#include <fstream>
+#include <memory>
+#include <string>
+
+#include "tools/cpp/runfiles/runfiles.h"
+
+using bazel::tools::cpp::runfiles::Runfiles;
+
+namespace {
+  std::unique_ptr<Runfiles> runfiles = nullptr;
```
stefanbucur, line 29: We follow Google's C++ style guide, which does not allow non-trivially destructible types for global variables: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

Here, given the static storage of this value, it is sufficient to store it as a global static pointer (it won't be a memory leak).

examples/runfiles_fuzz_test.cc
```diff
@@ -0,0 +1,44 @@
...
+#include <cstddef>
+#include <cstdint>
+
+#include <fstream>
+#include <memory>
+#include <string>
+
+#include "tools/cpp/runfiles/runfiles.h"
+
+using bazel::tools::cpp::runfiles::Runfiles;
```
stefanbucur, line 26: Style nit: Use the `::bazel::tools::...` form (https://abseil.io/tips/119).

examples/runfiles_fuzz_test.cc
```diff
@@ -0,0 +1,44 @@
...
+  std::unique_ptr<Runfiles> runfiles = nullptr;
+}
+
+extern "C" void LLVMFuzzerInitialize(int *argc, char ***argv) {
+  std::string error;
+  runfiles = std::unique_ptr<Runfiles>(Runfiles::Create((*argv)[0], &error));
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  if (runfiles == nullptr) abort();
```
stefanbucur, line 38: Consider moving this check inside the initializer, where you can also print the error string in case of errors.

fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -15,26 +15,47 @@
...
+    archive_inputs = binary_runfiles
+
+    runfiles_manifest = ctx.actions.declare_file(ctx.label.name + "_runfiles")
+    runfiles_manifest_content = "".join([
+        "{runfile_path} {real_path}\n".format(
+            real_path = runfile.path,
+            runfile_path = runfile_path(ctx, runfile),
+        )
+        for runfile in binary_runfiles
+        if runfile != binary_info.binary_file
```
stefanbucur, line 34: This is an important detail that might be easy to miss: Consider adding a comment here explaining why we skip the binary file in the iteration.

stefanbucur (commented):
This looks great overall. Leaving a few smaller comments.

fmeum:
Thanks for the review of the C++ part. I pushed a new commit to address your comments.

stefanbucur (approved):

PR #150, opened 2021-05-17
Manually include <limits> to fix CI failures

fmeum:
This works around the CI failures observed in #146 through #149, but does not make any attempts to address the unknown root cause.

### DIFF
--- a/examples/fuzzed_data_provider_fuzz_test.cc
+++ b/examples/fuzzed_data_provider_fuzz_test.cc
@@ -14,10 +14,16 @@
 
 // A fuzz target that demonstrates the use of FuzzeddataProvider.
 
-#include <fuzzer/FuzzedDataProvider.h>
-
 #include <cstddef>
 #include <cstdint>
+
+// Workaround for
+// https://github.com/llvm/llvm-project/commit/0cccccf0d2cbd707503263785f9a0407d3e2bd5ea
+// causing missing symbol errors in the FuzzedDataProvider.h header with
+// clang 10.
+// FIXME: Remove once a clang release ships with this commit.
+#include <limits>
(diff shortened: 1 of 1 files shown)


examples/fuzzed_data_provider_fuzz_test.cc
```diff
@@ -14,11 +14,14 @@
 
 // A fuzz target that demonstrates the use of FuzzeddataProvider.
 
-#include <fuzzer/FuzzedDataProvider.h>
-
 #include <cstddef>
 #include <cstdint>
 
+// Work around a CI failure due to symbols from <limits> missing in the
+// FuzzedDataProvider.h header.
```
stefanbucur, line 21: Maybe mention here that we should remove this header here once we update to a Clang version more recent than https://github.com/llvm/llvm-project/commit/0cccccf0d2cbd707503263785f9a0407d3e2bd5e?

stefanbucur:
Thanks for fixing this issue! This looks indeed like an LLVM bug that got fixed in https://github.com/llvm/llvm-project/commit/0cccccf0d2cbd707503263785f9a0407d3e2bd5e

stefanbucur (approved):

examples/fuzzed_data_provider_fuzz_test.cc
fmeum: Thanks for finding the root cause, done. Do you understand why this only became an issue now?

stefanbucur: > Do you understand why this only became an issue now?

That's a really good question. Unfortunately I can't think of a concrete good reason - the builds are supposed to be hermetic and deterministic. I suspect either changes in the Bazel version or the underlying Ubuntu image might have caused a change in compilation order, which in turn exposed this missing include.

PR #151, opened 2021-05-17
Use OSS-Fuzz provided top-level files

fmeum:
OSS-Fuzz provides certain files for use by all fuzz targets depending on the language declared in the project.yaml, e.g. the Jazzer driver and agent. These files should be used instead of building them themselves, primarily because this allows seemless updates to the fuzzing engines without requiring the project to update to a new version of rules_fuzzing.

This commit introduces a list of runfiles (identified by workspace name and short_path) that should be replaced with the OSS-Fuzz provided files. This is achieved by replacing the actual runfile in the generated tar archive with a relative symlink pointing to the corresponding file with the same basename in the top-level $OUT directory.

This resolves 1. in https://github.com/bazelbuild/rules_fuzzing/pull/146#discussion_r632299433 in a generic way that can be adapted easily to cover other languages.

Depends on #149.

### DIFF
--- a/fuzzing/private/oss_fuzz/repository.bzl
+++ b/fuzzing/private/oss_fuzz/repository.bzl
@@ -103,12 +103,37 @@ def _extract_build_params(
         instrum_cxxopts = instrum_cxxopts,
     )
 
+# The paths under $OUT of the OSS-Fuzz provided Jazzer files.
+_JAZZER_IN_AGENT_DEPLOY_JAR = "jazzer_agent_deploy.jar"
+_JAZZER_IN_DRIVER = "jazzer_driver"
+_JAZZER_IN_DRIVER_ADDRESS_SANITIZER = "jazzer_driver_asan"
+
+# The filenames under which the OSS-Fuzz provided Jazzer files will be available
+# in this repository.
+_JAZZER_OUT_AGENT_DEPLOY_JAR = _JAZZER_IN_AGENT_DEPLOY_JAR
+_JAZZER_OUT_DRIVER_NO_SANITIZER = "jazzer_driver_no_sanitizer"
+_JAZZER_OUT_DRIVER_WITH_SANITIZER = "jazzer_driver_with_sanitizer"
+
+def _export_jazzer(repository_ctx, out_path, sanitizer):
+    if out_path == None:
+        return []
+    exported_files = []
+    repository_ctx.symlink(out_path + "/" + _JAZZER_IN_AGENT_DEPLOY_JAR, _JAZZER_OUT_AGENT_DEPLOY_JAR)
+    exported_files.append(_JAZZER_OUT_AGENT_DEPLOY_JAR)
+    repository_ctx.symlink(out_path + "/" + _JAZZER_IN_DRIVER, _JAZZER_OUT_DRIVER_NO_SANITIZER)
+    exported_files.append(_JAZZER_OUT_DRIVER_NO_SANITIZER)
+    if sanitizer == "address":
+        repository_ctx.symlink(out_path + "/" + _JAZZER_IN_DRIVER_ADDRESS_SANITIZER, _JAZZER_OUT_DRIVER_WITH_SANITIZER)
+        exported_files.append(_JAZZER_OUT_DRIVER_WITH_SANITIZER)
+    return exported_files
+
 def _oss_fuzz_repository(repository_ctx):
     environ = repository_ctx.os.environ
     fuzzing_engine_library = environ.get("LIB_FUZZING_ENGINE")
     sanitizer = environ.get("SANITIZER")
     cflags = environ.get("FUZZING_CFLAGS") or environ.get("CFLAGS", "")
     cxxflags = environ.get("FUZZING_CXXFLAGS") or environ.get("CXXFLAGS", "")
+    out_path = environ.get("OUT")
 
     build_params = _extract_build_params(
         repository_ctx,
@@ -117,6 +142,7 @@ def _oss_fuzz_repository(repository_ctx):
         cflags.split(" "),
         cxxflags.split(" "),
     )
+    exported_files = _export_jazzer(repository_ctx, out_path, sanitizer)
 
     repository_ctx.template(
         "BUILD",
@@ -124,6 +150,7 @@ def _oss_fuzz_repository(repository_ctx):
         {
             "%{stub_srcs}": _to_list_repr(build_params.stub_srcs),
             "%{stub_linkopts}": _to_list_repr(build_params.stub_linkopts),
+            "%{exported_files}": _to_list_repr(exported_files),
         },
     )
     repository_ctx.template(
@@ -148,6 +175,7 @@ oss_fuzz_repository = repository_rule(
         "CFLAGS",
         "CXXFLAGS",
         "SANITIZER",
+        "OUT",
     ],
     local = True,
     doc = """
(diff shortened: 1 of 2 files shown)


fuzzing/private/oss_fuzz/package.bzl
```diff
@@ -44,6 +44,32 @@ def _oss_fuzz_package_impl(ctx):
...
+                jazzer/driver/jazzer_driver_asan \
+                "
+            for RUNFILE in $RUNFILES_TO_PATCH; do
+                RUNFILE_PATH="{base_name}.runfiles/$RUNFILE"
+                [[ -e "$STAGING_DIR/$RUNFILE_PATH" ]] || continue
+                tar --delete -f "{output}" "./$RUNFILE_PATH"
+                # Generate a path that when resolved relative to RUNFILE_PATH's
+                # parent directory points to a top-level OSS-Fuzz binary of the
+                # same basename.
+                RUNFILE_RELATIVE_PATH_TO_GLOBAL="$(echo $RUNFILE_PATH | sed 's|[^/]*/|../|g')"
```
stefanbucur, line 63: What is a top-level OSS-Fuzz binary? Is that an absolute path?

I'm not sure such paths would actually be available to the Bazel rule during building, since all builds are hermetic by default and can only access the explicitly-mentioned input files.

stefanbucur (commented):
I am concerned that this change transforms the packaging rule from a purely hermetic, no-assumptions rule, to one that gets coupled to a particular language and a particular layout of OSS-Fuzz files.

I would like to discuss design alternatives that preserve the separation of concerns in the system. In this case, the packaging rule should be concerned with assembling the fuzz test resources in a language-agnostic way.

For C++, we have created a special repository rule that generated a C++ fuzzing engine based off the files available in the OSS-Fuzz environment. Does the need to "patch" Java fuzz targets constitute a sign that this might actually be needed here, too?

fmeum:
> I am concerned that this change transforms the packaging rule from a purely hermetic, no-assumptions rule, to one that gets coupled to a particular language and a particular layout of OSS-Fuzz files.

I share your concern and would also like to achieve this goal.

> For C++, we have created a special repository rule that generated a C++ fuzzing engine based off the files available in the OSS-Fuzz environment. Does the need to "patch" Java fuzz targets constitute a sign that this might actually be needed here, too?

This is a very good idea and I will look into whether this is possible. It could turn out to be a much better solution.

fuzzing/private/oss_fuzz/package.bzl
fmeum: What I mean by "top-level OSS-Fuzz binary" here is a binary made available under a fixed name directly in `$OUT/` by OSS-Fuzz `compile` script. This currently applies to `jazzer_driver`, `jazzer_driver_asan`, and `jazzer_agent_deploy.jar` if the language of the project is set to `jvm`.

These paths would not need to be available to Bazel during building the way this PR is currently designed: In the final step, the actual runfile version of those files built with Bazel are replaced by relative symlinks that resolve to the corresponding binaries in `$OUT/`. Since the resulting .tar archives are extracted directly into `$OUT/`, the relative path from the runfile `jazzer_driver` to `$OUT/jazzer_driver` is fixed and can be computed up front, which is what this line is supposed to do. Since the symlink is dangling while in the tar and never resolved (note the lack of `-h` in the last `tar` command), this works fine in the sandbox.

I'm not sure I don't miss anything here, but I believe the only assumption this part of the script makes is that the target `@jazzer//driver:jazzer_driver` in any Bazel project that uses `rules_fuzzing` should always be replaced with the OSS-Fuzz provided `jazzer_driver` (and e.g. is not something else entirely that just happens to have this label).

fmeum:
I tried out doing this with a repository rule and this works well with one caveat: Without any special casing, the files in `$OUT` will be copied into the runfiles tree rather than symlinked into it. I don't think this would break anything as these binaries do not depend on being in a particular location, but it does deviate from how they are usually used and duplicates them for every fuzz target. I lean towards the repository rule as it is cleaner, but am not entirely sure about the tradeoff.

stefanbucur:
> I tried out doing this with a repository rule and this works well with one caveat: Without any special casing, the files in `$OUT` will be copied into the runfiles tree rather than symlinked into it. I don't think this would break anything as these binaries do not depend on being in a particular location, but it does deviate from how they are usually used and duplicates them for every fuzz target. I lean towards the repository rule as it is cleaner, but am not entirely sure about the tradeoff.

I also prefer the repository rule, as it is conceptually cleaner: The packaging rule ends up not being coupled with anything else, so it is simpler to reason about and maintain going forward. There are also fewer hidden dependencies between system components.

Do you have a sense of how large the duplicated files are? If they are only on the order of MBs, that should not matter much IMO.

fmeum:
> Do you have a sense of how large the duplicated files are? If they are only on the order of MBs, that should not matter much IMO.

They total ~15MB per target. Since Jazzer itself does not care about its own path and the `llvm-symbolizer` most likely doesn't as well, I will redo this PR in the form of a repository rule.

stefanbucur:
> > Do you have a sense of how large the duplicated files are? If they are only on the order of MBs, that should not matter much IMO.
> 
> They total ~15MB per target. Since Jazzer itself does not care about its own path and the `llvm-symbolizer` most likely doesn't as well, I will redo this PR in the form of a repository rule.

Sounds good to me! Thank you.

fmeum:
@stefanbucur I have pushed a new commit that exports the relevant files from `@rules_fuzzing_oss_fuzz`. Once this has been merged, I can update #146 to use these files.

fuzzing/private/oss_fuzz/repository.bzl
```diff
@@ -103,12 +103,37 @@ def _extract_build_params(
...
+# The paths under $OUT of the OSS-Fuzz provided Jazzer files.
+_JAZZER_IN_AGENT_DEPLOY_JAR = "jazzer_agent_deploy.jar"
+_JAZZER_IN_DRIVER = "jazzer_driver"
+_JAZZER_IN_DRIVER_ADDRESS_SANITIZER = "jazzer_driver_asan"
+
+# The filenames under which the OSS-Fuzz provided Jazzer files will be available
+# in this repository.
+_JAZZER_OUT_AGENT_DEPLOY_JAR = _JAZZER_IN_AGENT_DEPLOY_JAR
+_JAZZER_OUT_DRIVER_NO_SANITIZER = "jazzer_driver_no_sanitizer"
+_JAZZER_OUT_DRIVER_WITH_SANITIZER = "jazzer_driver_with_sanitizer"
```
stefanbucur, line 115: If there is an intuitive connection between the `*_IN_*` and `*_OUT_*` forms, I think it would be more visible if the variables were grouped by name, e.g.:

```
_JAZZER_IN_AGENT_DEPLOY_JAR = ...
_JAZZER_OUT_AGENT_DEPLOY_JAR = ...

_JAZZER_IN_DRIVER = ...
_JAZZER_IN_DRIVER_ADDRESS_SANITIZER = ...
_JAZZER_OUT_DRIVER_NO_SANITIZER = ...
_JAZZER_OUT_DRIVER_WITH_SANITIZER = ...
```

stefanbucur (approved):

fuzzing/private/oss_fuzz/repository.bzl
fmeum: I grouped them as you suggested, but also added further comments. Let me know if you find this too verbose.

stefanbucur: These are useful, thanks!

PR #152, opened 2021-06-10
Account for new OSS-Fuzz Jazzer binary names

fmeum:
Since https://github.com/google/oss-fuzz/commit/73d78b88790b501f119801c4f68463180b76e1d9, OSS-Fuzz now provides the Jazzer driver with the correct sanitizer in jazzer_driver_with_sanitizer. This simplifies the logic in `repository.bzl`.


stefanbucur (approved):
Great!

PR #153, opened 2021-06-15
Various tweaks in engine launchers for regression mode.

xinhaoyuan:
For LibFuzzer, do not read the output corpus in regression mode.

For Honggfuzz, use dry-run-only mode instead of minimize mode for more expected behavior, and enable outputs from the target for easier debugging.


stefanbucur (approved):
Thanks for these fixes!

PR #157, opened 2021-06-20
Add OSS-Fuzz support to java_fuzz_test

fmeum:
This commit makes the <basename>_oss_fuzz targets of java_fuzz_test
instances work in OSS-Fuzz, based on the runfiles support and the
handling of OSS-Fuzz provided binaries introduced in previous PRs.

The following parts of the commit are notable:
* The Jazzer update results in a simplified runfiles structure of the
  driver, which prevents runfile paths clashes.
* Since .options files are discouraged, target-dependant shell scripts
  are used to set the required options for sanitizers.
* The transitive_native_deps attribute of java_fuzz_test is removed in
  favor of an overapproximation of the Java library path, which makes
  the java_fuzz_test easier to use with Bazel 4.

Replaces #146.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,8 @@ Our libFuzzer test will start running and immediately discover the buffer overfl
 
 ```
 INFO: Seed: 2957541205
-INFO: Loaded 1 modules   (8 inline 8-bit counters): 8 [0x5aab10, 0x5aab18), 
-INFO: Loaded 1 PC tables (8 PCs): 8 [0x5aab18,0x5aab98), 
+INFO: Loaded 1 modules   (8 inline 8-bit counters): 8 [0x5aab10, 0x5aab18),
+INFO: Loaded 1 PC tables (8 PCs): 8 [0x5aab18,0x5aab98),
 INFO:      755 files found in /tmp/fuzzing/corpus
 INFO:        0 files found in fuzz_test_corpus
 INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 35982 bytes
@@ -214,7 +214,7 @@ INFO: seed corpus: files: 16 min: 1b max: 19b total: 210b rss: 199Mb
 
 Once you wrote and tested the fuzz test, you should run it on continuous fuzzing infrastructure so it starts generating tests and finding new crashes in your code.
 
-The C++ fuzzing rules provide out-of-the-box support for [OSS-Fuzz](https://github.com/google/oss-fuzz), free continuous fuzzing infrastructure from Google for open source projects. Read its [Bazel project guide][bazel-oss-fuzz] for detailed instructions.
+The fuzzing rules provide out-of-the-box support for [OSS-Fuzz](https://github.com/google/oss-fuzz), free continuous fuzzing infrastructure from Google for open source projects. Read its [Bazel project guide][bazel-oss-fuzz] for detailed instructions.
 
 ## Where to go from here?
 
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ INFO: seed corpus: files: 16 min: 1b max: 19b total: 210b rss: 199Mb
 
 Once you wrote and tested the fuzz test, you should run it on continuous fuzzing infrastructure so it starts generating tests and finding new crashes in your code.
 
-The fuzzing rules provide out-of-the-box support for [OSS-Fuzz](https://github.com/google/oss-fuzz), free continuous fuzzing infrastructure from Google for open source projects. Read its [Bazel project guide][bazel-oss-fuzz] for detailed instructions.
+The C++ and Java fuzzing rules provide out-of-the-box support for [OSS-Fuzz](https://github.com/google/oss-fuzz), free continuous fuzzing infrastructure from Google for open source projects. Read its [Bazel project guide][bazel-oss-fuzz] for detailed instructions.
 
 ## Where to go from here?
 
--- a/examples/java/BUILD
+++ b/examples/java/BUILD
@@ -38,6 +38,9 @@ java_fuzz_test(
 java_fuzz_test(
     name = "JavaFuzzTest",
     srcs = ["com/example/JavaFuzzTest.java"],
+    tags = [
+        "no-oss-fuzz",
+    ],
 )
 
 java_fuzz_test(
@@ -46,7 +49,10 @@ java_fuzz_test(
     corpus = [
         ":corpus",
     ],
-    transitive_native_deps = [
+    tags = [
+        "no-oss-fuzz",
+    ],
+    deps = [
         ":native",
     ],
 )
--- a/examples/java/BUILD
+++ b/examples/java/BUILD
@@ -57,6 +57,18 @@ java_fuzz_test(
     ],
 )
 
+java_fuzz_test(
+    name = "JavaNativeRunfileFuzzTest",
+    srcs = ["com/example/JavaNativeRunfileFuzzTest.java"],
+    data = [
+        "corpus_0.txt",
+    ],
+    deps = [
+        ":native_runfile",
+        "@bazel_tools//tools/java/runfiles",
+    ],
+)
+
 # A native library that interfaces with Java through the JNI.
 cc_binary(
     name = "native",
@@ -66,6 +78,22 @@ cc_binary(
     ],
     linkshared = True,
     deps = [
+        "@bazel_tools//tools/jdk:jni",
+    ],
+)
+
+cc_binary(
+    name = "native_runfile",
+    srcs = [
+        "com/example/JavaNativeRunfileFuzzTest.cpp",
+        "com/example/JavaNativeRunfileFuzzTest.h",
+    ],
+    data = [
+        "corpus_1.txt",
+    ],
+    linkshared = True,
+    deps = [
+        "@bazel_tools//tools/cpp/runfiles",
         "@bazel_tools//tools/jdk:jni",
     ],
 )
--- a/examples/java/BUILD
+++ b/examples/java/BUILD
@@ -91,6 +91,8 @@ cc_binary(
     data = [
         "corpus_1.txt",
     ],
+    # Build as a shared library that can be loaded by a Java application at
+    # runtime via System.loadLibrary().
     linkshared = True,
     deps = [
         "@bazel_tools//tools/cpp/runfiles",
--- a/fuzzing/private/java_utils.bzl
+++ b/fuzzing/private/java_utils.bzl
@@ -15,6 +15,7 @@
 """Utilities and helper rules for Java fuzz tests."""
 
 load("//fuzzing/private:binary.bzl", "fuzzing_binary_transition")
+load("//fuzzing/private:util.bzl", "runfile_path")
 
 # A Starlark reimplementation of a part of Bazel's JavaCommon#determinePrimaryClass.
 def determine_primary_class(srcs, name):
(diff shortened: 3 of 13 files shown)


fmeum:
@stefanbucur What is your preferred way to prevent the Java fuzz tests from running in the existing CIFuzz project? I will set up a separate test project.

fmeum:
@stefanbucur The CI failure is very unexpected and already reproduces on current `master` as follows:
```
$ git clone https://github.com/google/oss-fuzz
$ cd oss-fuzz
$ infra/helper.py build_image bazel-rules-fuzzing-test
Pull latest base images (compiler/runtime)? (y/N): y
Pulling latest base images...
Running: docker pull gcr.io/oss-fuzz-base/base-image
Using default tag: latest
latest: Pulling from oss-fuzz-base/base-image
Digest: sha256:a882198e762e06a0f722dbb585c6ab73f4e4284ba8950ac1b27d7d7abcc5252a
Status: Image is up to date for gcr.io/oss-fuzz-base/base-image:latest
gcr.io/oss-fuzz-base/base-image:latest
Running: docker pull gcr.io/oss-fuzz-base/base-clang
Using default tag: latest
latest: Pulling from oss-fuzz-base/base-clang
Digest: sha256:d34cefbc31613bf6f8edcb4af0738d5344c1cb3700fa3c21adbc7cb8ebf7bcf0
Status: Image is up to date for gcr.io/oss-fuzz-base/base-clang:latest
gcr.io/oss-fuzz-base/base-clang:latest
Running: docker pull gcr.io/oss-fuzz-base/base-builder
Using default tag: latest
latest: Pulling from oss-fuzz-base/base-builder
Digest: sha256:bba0207d3012b7d566c98c17fc92c384a6f54cf71b0b28576ba98f97efd8d9fa
Status: Image is up to date for gcr.io/oss-fuzz-base/base-builder:latest
gcr.io/oss-fuzz-base/base-builder:latest
Running: docker pull gcr.io/oss-fuzz-base/base-runner
Using default tag: latest
latest: Pulling from oss-fuzz-base/base-runner
Digest: sha256:523633f5358523baff0cc2ec7f7717072eac9316643130c86d953ef2fb902fde
Status: Image is up to date for gcr.io/oss-fuzz-base/base-runner:latest
gcr.io/oss-fuzz-base/base-runner:latest
Running: docker pull gcr.io/oss-fuzz-base/base-runner-debug
Using default tag: latest
latest: Pulling from oss-fuzz-base/base-runner-debug
Digest: sha256:2b4b4105a9ce5e9888787a10e288f4d8bcc702511696fc786c40fc4f235878a3
Status: Image is up to date for gcr.io/oss-fuzz-base/base-runner-debug:latest
gcr.io/oss-fuzz-base/base-runner-debug:latest
Running: docker pull gcr.io/oss-fuzz-base/base-sanitizer-libs-builder
Using default tag: latest
latest: Pulling from oss-fuzz-base/base-sanitizer-libs-builder
Digest: sha256:bb6bb6b5072332a0f0ed8d3e6c50859bdcfff22b497ff025084e6bf6b8e1f397
Status: Image is up to date for gcr.io/oss-fuzz-base/base-sanitizer-libs-builder:latest
gcr.io/oss-fuzz-base/base-sanitizer-libs-builder:latest
Running: docker pull gcr.io/oss-fuzz-base/msan-libs-builder
Using default tag: latest
latest: Pulling from oss-fuzz-base/msan-libs-builder
Digest: sha256:8f2cc9c5eb36afb4d6b163490cdd2f48d977a909d70b59264df374d326a78cd6
Status: Image is up to date for gcr.io/oss-fuzz-base/msan-libs-builder:latest
gcr.io/oss-fuzz-base/msan-libs-builder:latest
Running: docker build --no-cache -t gcr.io/oss-fuzz/bazel-rules-fuzzing-test projects/bazel-rules-fuzzing-test
Sending build context to Docker daemon  5.632kB
Step 1/4 : FROM gcr.io/oss-fuzz-base/base-builder
 ---> d9cb8aaba0c9
Step 2/4 : RUN git clone https://github.com/bazelbuild/rules_fuzzing.git
 ---> Running in eb6ff9ad0e13
Cloning into 'rules_fuzzing'...
Removing intermediate container eb6ff9ad0e13
 ---> 42507c2e6ec2
Step 3/4 : WORKDIR $SRC/rules_fuzzing/
 ---> Running in 7d9a8d1a8efd
Removing intermediate container 7d9a8d1a8efd
 ---> 51c4cf8f459c
Step 4/4 : COPY build.sh $SRC/
 ---> 89259c35b862
Successfully built 89259c35b862
Successfully tagged gcr.io/oss-fuzz/bazel-rules-fuzzing-test:latest
$ infra/helper.py shell bazel-rules-fuzzing-test
$ bazel build //... --action_env=CC=${CC} --action_env=CXX=${CXX}
2021/06/22 13:33:03 Downloading https://releases.bazel.build/4.1.0/release/bazel-4.1.0-linux-x86_64...

Extracting Bazel installation...
Starting local Bazel server and connecting to it...
INFO: Analyzed 82 targets (94 packages loaded, 2160 targets configured).
INFO: Found 82 targets...
ERROR: /root/.cache/bazel/_bazel_root/17db7bb59affe64429348778697da03a/external/jazzer/driver/BUILD.bazel:21:11: Compiling driver/fuzzed_data_provider.cpp failed: (Exit 1): clang failed: error executing command /usr/local/bin/clang -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wno-free-nonheap-object -fcolor-diagnostics -fno-omit-frame-pointer '-std=c++0x' -MD -MF ... (remaining 55 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
In file included from external/jazzer/driver/fuzzed_data_provider.cpp:56:
In file included from external/com_google_absl/absl/strings/str_format.h:76:
In file included from external/com_google_absl/absl/strings/internal/str_format/arg.h:30:
external/com_google_absl/absl/meta/type_traits.h:620:37: error: unknown template name 'invoke_result'
struct result_of<F(Args...)> : std::invoke_result<F, Args...> {};
                                    ^
1 error generated.
INFO: Elapsed time: 149.887s, Critical Path: 24.91s
INFO: 245 processes: 111 internal, 123 linux-sandbox, 11 worker.
FAILED: Build did NOT complete successfully
```

The build passes locally and it passed in the CI when we merged the other PRs.

fmeum:
This is fixed by https://github.com/google/oss-fuzz/pull/5960/commits/1d3c4555d9c614d7d785c827514a6e6df894471f. Without that commit, uninstrumented C++ binaries (such as the Jazzer driver) will be built against the system C++ standard library, which is not the one included with the clang built from source and likely is too old for C++17 support.

examples/java/BUILD
```diff
@@ -63,3 +81,19 @@ cc_binary(
...
+cc_binary(
+    name = "native_runfile",
+    srcs = [
+        "com/example/JavaNativeRunfileFuzzTest.cpp",
+        "com/example/JavaNativeRunfileFuzzTest.h",
+    ],
+    data = [
+        "corpus_1.txt",
+    ],
+    linkshared = True,
```
stefanbucur, line 94: Since this attribute appears unusual, perhaps add a small comment explaining why it is needed?

fuzzing/private/java_utils.bzl
```diff
@@ -80,51 +81,65 @@ def _java_segment_index(path_segments):
...
-    --jvm_args="-Djava.library.path={library_path}" \
+    # The script is split into two parts: The first is emitted as-is, the second
+    # is a template that is passed to format(). Without the split, curly braces
+    # in the first part would need to be escaped.
+    script_literal_part = """#!/bin/bash
+# LLVMFuzzerTestOneInput - OSS-Fuzz needs this string literal to appear
+# somewhere in the script so it is recognized as a fuzz target.
+
+# --- begin runfiles.bash initialization v2 ---
+# Copy-pasted from the Bazel Bash runfiles library v2.
```
stefanbucur, line 95: Could we reuse the library here, instead of copy-pasting it?

fmeum: That is unfortunately not possible, see https://github.com/bazelbuild/bazel/blob/e3c4f4369c3c37d14d95146b69e49386290a988f/tools/bash/runfiles/runfiles.bash#L50. We have a chicken-and-egg problem where we first need to find the library, which is itself a runfile, before we can use it to find runfiles.

examples/java/BUILD
fmeum: I added a comment.

README.md
```diff
@@ -214,7 +214,7 @@ INFO: seed corpus: files: 16 min: 1b max: 19b total: 210b rss: 199Mb
 
 Once you wrote and tested the fuzz test, you should run it on continuous fuzzing infrastructure so it starts generating tests and finding new crashes in your code.
 
-The C++ fuzzing rules provide out-of-the-box support for [OSS-Fuzz](https://github.com/google/oss-fuzz), free continuous fuzzing infrastructure from Google for open source projects. Read its [Bazel project guide][bazel-oss-fuzz] for detailed instructions.
+The fuzzing rules provide out-of-the-box support for [OSS-Fuzz](https://github.com/google/oss-fuzz), free continuous fuzzing infrastructure from Google for open source projects. Read its [Bazel project guide][bazel-oss-fuzz] for detailed instructions.
```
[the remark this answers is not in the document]
fmeum: Sounds good!

examples/java/BUILD
stefanbucur: Thanks!

fuzzing/private/java_utils.bzl
stefanbucur: Oh, I see! Could you leave a comment in this respect? E.g. "Bazel-provided code snippet that should be copy-pasted as is at use sites."

stefanbucur:
> @stefanbucur What is your preferred way to prevent the Java fuzz tests from running in the existing CIFuzz project? I will set up a separate test project.

Really good point. I can think of a number of options:

* Define separate tags for each fuzz test language (e.g., `cc-fuzz-test` and `java-fuzz-test`) and set a filter in the bazel query on the OSS-Fuzz side. The project owner would have to explicitly set this filter in the build.sh file.
* Automatically deduct the fuzz test type based on the configured project language.

I remember from past discussions that the language attribute of the project was causing trouble for some projects. I can imagine how this could be problematic for Bazel projects, which tend to naturally allow multiple languages in direct and transitive deps.

So maybe we can go with explicitly filtering the tests of interest in the build.sh file?

fuzzing/private/java_utils.bzl
fmeum: Sounds good!

fmeum:
> > @stefanbucur What is your preferred way to prevent the Java fuzz tests from running in the existing CIFuzz project? I will set up a separate test project.
> 
> Really good point. I can think of a number of options:
> 
> * Define separate tags for each fuzz test language (e.g., `cc-fuzz-test` and `java-fuzz-test`) and set a filter in the bazel query on the OSS-Fuzz side. The project owner would have to explicitly set this filter in the build.sh file.
> * Automatically deduct the fuzz test type based on the configured project language.
> 
> I remember from past discussions that the language attribute of the project was causing trouble for some projects. I can imagine how this could be problematic for Bazel projects, which tend to naturally allow multiple languages in direct and transitive deps.
> 
> So maybe we can go with explicitly filtering the tests of interest in the build.sh file?

This has been resolved in one way by https://github.com/google/oss-fuzz/pull/5960/commits/3954b7a20d1f026d5e2ca57d1b9541547a82370a, leaning on the `language` setting in the project YAML. This results in filtering out the fuzz tests for other languages.

Long term, if e.g. every fuzz target indicated its own language somehow and multi-language projects were possible on OSS-Fuzz, we could just drop the filtering and run all fuzz tests in a single project.

fmeum:
@stefanbucur All CI checks are passing now that the docker images have been updated.

In https://github.com/bazelbuild/rules_fuzzing/runs/2853369174?check_suite_focus=true#step:5:17, you can see that even on current `master` CI Fuzz tries to run the `raw_` targets, the same happens in the CI runs for this PR. This is not a big problem, but may still warrant fixing. It is not related to the current PR though.

stefanbucur:
> @stefanbucur All CI checks are passing now that the docker images have been updated.
> 
> In https://github.com/bazelbuild/rules_fuzzing/runs/2853369174?check_suite_focus=true#step:5:17, you can see that even on current `master` CI Fuzz tries to run the `raw_` targets, the same happens in the CI runs for this PR. This is not a big problem, but may still warrant fixing. It is not related to the current PR though.

Does this mean OSS-Fuzz detects the `raw_` targets and attempts to run them in continuous fuzzing, too? That might actually be a problem, at least because it would make things confusing for developers.

stefanbucur (approved):

fmeum:
I have verified that is not the case and asked about the CI Fuzz behavior here: https://github.com/google/oss-fuzz/issues/5777#issuecomment-871939044.

stefanbucur:
> I have verified that is not the case and asked about the CI Fuzz behavior here: [google/oss-fuzz#5777 (comment)](https://github.com/google/oss-fuzz/issues/5777#issuecomment-871939044).

My reading of Jonathan's answer is different. He explicitly says "If your fuzzer build has fuzzers in subdirectories, OSS-Fuzz/ClusterFuzz will run them." This means that right now Bazel builds would run twice on OSS-Fuzz because of this, which wastes resources and it is suboptimal.

If this is the case, this is a high-priority issue that should be addressed. 

Is this a problem just for C++ targets? Maybe we can disable runfiles for those for the time being.

fmeum:
> > I have verified that is not the case and asked about the CI Fuzz behavior here: [google/oss-fuzz#5777 (comment)](https://github.com/google/oss-fuzz/issues/5777#issuecomment-871939044).
> 
> My reading of Jonathan's answer is different. He explicitly says "If your fuzzer build has fuzzers in subdirectories, OSS-Fuzz/ClusterFuzz will run them." This means that right now Bazel builds would run twice on OSS-Fuzz because of this, which wastes resources and it is suboptimal.
> 
> If this is the case, this is a high-priority issue that should be addressed.
> 
> Is this a problem just for C++ targets? Maybe we can disable runfiles for those for the time being.

I tested this with `infra/helper.py`, but I now see that this uses entirely different logic to detect what a fuzz target is. The relevant script is https://github.com/google/oss-fuzz/blob/beee192ed30387c315b412e517e71eb60d2d7dae/infra/base-images/base-runner/targets_list#L3, so yes, runfiles will mean everything is built twice.

A simple upstream fix could be to exclude symlinks on the `find` command. What do you think?

PR #158, opened 2021-06-27
Use correct `cc_engine_instrumentation` in `jazzer` config

fmeum:
While the `cc_engine_instrumentation` settings `libfuzzer` and `jazzer` lead to the same instrumentation, this may not remain so indefinitely. Thus the `jazzer` and `asan-jazzer` config in `.bazelrc` should use the `jazzer` value for the setting.

Also, even when not using a sanitizer via the `jazzer` config, native library dependencies should be instrumented for fuzzing.

This requires updating Jazzer to the latest version as previous versions did not correctly export certain sanitizer symbols.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -59,14 +59,14 @@ build:oss-fuzz --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
 
 # Jazzer without sanitizer (Java only)
 build:jazzer --//fuzzing:java_engine=//fuzzing/engines:jazzer
-build:jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=none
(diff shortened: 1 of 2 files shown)


fmeum:
Not sure why this issue has only appeared now, but it will be resolved by the extended logic for populating `LD_LIBRARY_PATH` in #157. This one can wait until the PR has been merged.

stefanbucur:
> Not sure why this issue has only appeared now, but it will be resolved by the extended logic for populating `LD_LIBRARY_PATH` in #157. This one can wait until the PR has been merged.

This PR looks good to me, will approve and merge once #157 lands.

fmeum:
The checks pass now.

stefanbucur (approved):

PR #161, opened 2021-07-03
Do not link to binary from its own runfiles tree

fmeum:
This deviates from the usual Bazel runfiles layout, but is required
since ClusterFuzz executes fuzz targets in subdirectories and would thus
duplicate every C++ fuzz target.


fmeum:
@stefanbucur With this and #159 and #160, I would say that Java fuzzing should be ready for a release.

stefanbucur (approved):
Thanks!

stefanbucur:
> @stefanbucur With this and #159 and #160, I would say that Java fuzzing should be ready for a release.

Awesome - thanks for preparing this so thoroughly! I will set aside some time to cut the release.

PR #165, opened 2021-07-09
Clarify role of unqualified flags in .bazelrc

fmeum:
These flags are only needed within rules_fuzzing itself to work around
limitations in Bazel 4 and should not be copied by users.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -61,12 +61,14 @@ build:oss-fuzz --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
 build:jazzer --//fuzzing:java_engine=//fuzzing/engines:jazzer
 build:jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=jazzer
 build:jazzer --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
-# Workaround for https://github.com/bazelbuild/bazel/issues/11128
-build:jazzer --//fuzzing:cc_engine_sanitizer=none
 
 # Jazzer + ASAN
 build:asan-jazzer --//fuzzing:java_engine=//fuzzing/engines:jazzer
 build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=jazzer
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -61,12 +61,14 @@ build:oss-fuzz --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
...
-# Workaround for https://github.com/bazelbuild/bazel/issues/11128
-build:jazzer --//fuzzing:cc_engine_sanitizer=none
 
 # Jazzer + ASAN
 build:asan-jazzer --//fuzzing:java_engine=//fuzzing/engines:jazzer
 build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=jazzer
 build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_sanitizer=asan
-# Workaround for https://github.com/bazelbuild/bazel/issues/11128
+
+# Internal workaround for https://github.com/bazelbuild/bazel/issues/11128.
```
stefanbucur, line 70: It is not clear here what exactly is the workaround. The use of these flags?

.bazelrc
```diff
@@ -61,12 +61,14 @@ build:oss-fuzz --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
 build:jazzer --//fuzzing:java_engine=//fuzzing/engines:jazzer
 build:jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=jazzer
 build:jazzer --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
-# Workaround for https://github.com/bazelbuild/bazel/issues/11128
-build:jazzer --//fuzzing:cc_engine_sanitizer=none
 
 # Jazzer + ASAN
 build:asan-jazzer --//fuzzing:java_engine=//fuzzing/engines:jazzer
```
stefanbucur, line 66: BTW, note that this is not copy-pasteable, either: `//fuzzing/engines:jazzer` must also be prefixed with `@rules_fuzzing` in user repos.

PR #166, opened 2021-07-10
Reorganize and update the Java documentation in the README file in a …

stefanbucur:
…single section. This should make it easier for readers to find all the Java-specific information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ build:asan-libfuzzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=libfuzz
 build:asan-libfuzzer --@rules_fuzzing//fuzzing:cc_engine_sanitizer=asan
 ```
 
-Examples for other combinations of fuzzing engine and sanitizer can be found in the project's [`.bazelrc`](/.bazelrc).
+Examples for other combinations of fuzzing engine and sanitizer can be found in the [User Guide](/docs/guide.md#configuring-the-bazelrc-file).
 
 ### Defining a C++ fuzz test
 
--- a/README.md
+++ b/README.md
@@ -166,8 +166,8 @@ To use Jazzer, it is convienient to also define a `.bazelrc` configuration, simi
 
 ```
 # Force the use of Clang for all builds (needed to build Jazzer).
-build --action_env=CC=clang-10
-build --action_env=CXX=clang++-10
+build --action_env=CC=clang
+build --action_env=CXX=clang++
 
 # Define --config=jazzer for Jazzer without sanitizer (Java only).
 build:jazzer --@rules_fuzzing//fuzzing:java_engine=@rules_fuzzing//fuzzing/engines:jazzer
--- a/README.md
+++ b/README.md
@@ -147,6 +147,8 @@ The crash is saved under `/tmp/fuzzing/artifacts` and can be further inspected.
 You can write `java_fuzz_tests` through the [Jazzer][jazzer-doc] fuzzing engine. You will need to enable it in your WORKSPACE `rules_fuzzing_dependencies` call:
 
 ```python
+load("@rules_fuzzing//fuzzing:repositories.bzl", "rules_fuzzing_dependencies")
+
 rules_fuzzing_dependencies(jazzer = True)
 
 load("@rules_fuzzing//fuzzing:init.bzl", "rules_fuzzing_init")
--- a/README.md
+++ b/README.md
@@ -4,12 +4,14 @@ This repository contains [Bazel](https://bazel.build/) [Starlark extensions](htt
 
 [Fuzzing](https://en.wikipedia.org/wiki/Fuzzing) is an effective technique for uncovering security and stability bugs in software. Fuzzing works by invoking the code under test (e.g., a library API) with automatically generated data, and observing its execution to discover incorrect behavior, such as memory corruption or failed invariants. Read more [here](https://github.com/google/fuzzing) about fuzzing, additional examples, best practices, and other resources.
 
+The rule library currently provides support for C++ and Java fuzz tests. Support for additional languages may be added in the future.
+
 ## Features at a glance
 
 * Multiple fuzzing engines out of the box:
   * [libFuzzer][libfuzzer-doc]
   * [Honggfuzz][honggfuzz-doc]
-  * [Jazzer][jazzer-doc]
+  * Java fuzzing through [Jazzer][jazzer-doc]
 * Multiple sanitizer configurations:
   * [Address Sanitizer][asan-doc]
   * [Memory Sanitizer][msan-doc]
@@ -22,8 +24,6 @@ This repository contains [Bazel](https://bazel.build/) [Starlark extensions](htt
 * Customization options:
   * Defining additional fuzzing engines
   * Customizing the behavior of the fuzz test rule.
-
-The rule library currently provides support for C++ and Java fuzz tests. Support for additional languages may be added in the future.
 
 Contributions are welcome! Please read the [contribution guidelines](/docs/contributing.md).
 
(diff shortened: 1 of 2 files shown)


stefanbucur:
@fmeum PTAL

README.md
```diff
@@ -152,11 +142,47 @@ INFO: seed corpus: files: 755 min: 1b max: 35982b total: 252654b rss: 35Mb
...
+load("@jazzer//:repositories.bzl", "jazzer_dependencies")
+
+jazzer_dependencies()
+
+load("@jazzer//:init.bzl", "jazzer_init")
+
+jazzer_init()
+```
+
+To use Jazzer, it is convienient to also define a `.bazelrc` configuration, similar to the C++ libFuzzer one above:
```
fmeum, line 165:
```suggestion
To use Jazzer, it is convenient to also define a `.bazelrc` configuration, similar to the C++ libFuzzer one above:
```

README.md
```diff
@@ -152,11 +142,47 @@ INFO: seed corpus: files: 755 min: 1b max: 35982b total: 252654b rss: 35Mb
 
 The crash is saved under `/tmp/fuzzing/artifacts` and can be further inspected.
 
-### Defining a Java fuzz test
+### Java fuzzing
+
+You can write `java_fuzz_tests` through the [Jazzer][jazzer-doc] fuzzing engine. You will need to enable it in your WORKSPACE `rules_fuzzing_dependencies` call:
+
+```python
+rules_fuzzing_dependencies(jazzer = True)
```
fmeum, line 150: It may be clearer to give the full working snippet here (excluding the `http_archive`), even if it repeats a few more lines. Otherwise the `rules_fuzzing_init()` call is a bit too easy to miss:

```python
load("@rules_fuzzing//fuzzing:repositories.bzl", "rules_fuzzing_dependencies")

rules_fuzzing_dependencies(jazzer = True)

load("@rules_fuzzing//fuzzing:init.bzl", "rules_fuzzing_init")

rules_fuzzing_init()

load("@jazzer//:repositories.bzl", "jazzer_dependencies")

jazzer_dependencies()

load("@jazzer//:init.bzl", "jazzer_init")

jazzer_init()
```

README.md
```diff
@@ -152,11 +142,47 @@ INFO: seed corpus: files: 755 min: 1b max: 35982b total: 252654b rss: 35Mb
...
+
+load("@jazzer//:init.bzl", "jazzer_init")
+
+jazzer_init()
+```
+
+To use Jazzer, it is convienient to also define a `.bazelrc` configuration, similar to the C++ libFuzzer one above:
+
+```
+# Force the use of Clang for all builds (needed to build Jazzer).
```
fmeum, line 168: Would you say that g++ compilability is desirable here? The [most recent Jazzer commit](https://github.com/CodeIntelligenceTesting/jazzer) achieves that and I could update the `repository.bzl` reference. Of course, since `-fsanitize=fuzzer-no-link` is not supported, this would just enable Java-only fuzzing.

fmeum (approved):
Thank you very much, this flow is indeed much easier to follow (and more correct)!

The C++ part of the README still contains the line:
```
Examples for other combinations of fuzzing engine and sanitizer can be found in the project's [`.bazelrc`](/.bazelrc).
```

This is why I thought that `.bazelrc` would indeed be a place users should look for configurations, which then ideally wouldn't need any tweaking. Do you think it would make sense to remove that line and instead link to a section towards the end of the README, where we could include a copy-pastable version of the configs in `.bazelrc`?

README.md
```diff
@@ -152,11 +142,47 @@ INFO: seed corpus: files: 755 min: 1b max: 35982b total: 252654b rss: 35Mb
...
+build --action_env=CC=clang-10
+build --action_env=CXX=clang++-10
```
fmeum, line 170: Fixing a particular version of `clang` may be problematic here (although it is of course the right thing to do in `rules_fuzzing`'s own `.bazelrc`). How about:

```suggestion
# Force the use of Clang for all builds (Jazzer requires at least clang 9).
build --action_env=CC=clang
build --action_env=CXX=clang++
```

README.md
```diff
@@ -152,11 +142,47 @@ INFO: seed corpus: files: 755 min: 1b max: 35982b total: 252654b rss: 35Mb
...
+# Define --config=asan-jazzer for Jazzer + ASAN.
+build:asan-jazzer --@rules_fuzzing//fuzzing:java_engine=@rules_fuzzing//fuzzing/engines:jazzer
+build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=jazzer
+build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_sanitizer=asan
+```
 
 A Java fuzz test is specified using a [`java_fuzz_test` rule](/docs/java-fuzzing-rules.md#java_fuzz_test). In the most basic form, a Java fuzz test consists of a single `.java` file with a class that defines a function `public static fuzzerTestOneInput(byte[] input)`.
 
-The Java equivalent of the C++ fuzz test above would look as follows:
+Create the `com/example/JavaFuzzTest.java` file in your workspace root, as follows:
```
fmeum, line 185: Following the Maven layout simplifies the `java_fuzz_test` target below:

```suggestion
Create the `src/com/example/JavaFuzzTest.java` file in your workspace root, as follows:
```

README.md
```diff
@@ -172,20 +198,20 @@ public class JavaFuzzTest {
...
 
 ```python
 load("@rules_fuzzing//fuzzing:java_defs.bzl", "java_fuzz_test")
 
 java_fuzz_test(
     name = "JavaFuzzTest",
-    srcs = ["JavaFuzzTest.java"],
+    srcs = ["com/example/JavaFuzzTest.java"],
     # target_class is not needed if using the Maven directory layout.
     target_class = "com.example.JavaFuzzTest",
```
fmeum, line 210: With the path change above, this allows to drop `target_class`:
```suggestion
    srcs = ["src/com/example/JavaFuzzTest.java"],
    # target_class is not needed if using the Maven directory layout.
    # target_class = "com.example.JavaFuzzTest",
```

stefanbucur:
> Thank you very much, this flow is indeed much easier to follow (and more correct)!
> 
> The C++ part of the README still contains the line:
> 
> ```
> Examples for other combinations of fuzzing engine and sanitizer can be found in the project's [`.bazelrc`](/.bazelrc).
> ```
> 
> This is why I thought that `.bazelrc` would indeed be a place users should look for configurations, which then ideally wouldn't need any tweaking. Do you think it would make sense to remove that line and instead link to a section towards the end of the README, where we could include a copy-pastable version of the configs in `.bazelrc`?

I'm not sure if our `.bazelrc` filr will end up including an exhaustive set of valid configurations in the long run. But I think it's a good idea to have the most common / useful ones as a copy-pasteable version in our docs. Perhaps not in the README, since it would get quite verbose, but in the user guide. Let me take a stab at it.

README.md
stefanbucur: Good catch - fixed.

README.md
stefanbucur: > Would you say that g++ compilability is desirable here?

I'm not really sure, as I'm not familiar with the typical toolchains used to build native code mixed with Java code. I guess we can consider adding support for it if it ends up breaking some of our users.

README.md
stefanbucur: Done.

README.md
stefanbucur: Done.

README.md
stefanbucur: Oops, good catch.

README.md
stefanbucur: Good call, done.

README.md
```diff
@@ -152,11 +142,47 @@ INFO: seed corpus: files: 755 min: 1b max: 35982b total: 252654b rss: 35Mb
 
 The crash is saved under `/tmp/fuzzing/artifacts` and can be further inspected.
 
-### Defining a Java fuzz test
+### Java fuzzing
+
+You can write `java_fuzz_tests` through the [Jazzer][jazzer-doc] fuzzing engine. You will need to enable it in your WORKSPACE `rules_fuzzing_dependencies` call:
```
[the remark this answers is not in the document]
stefanbucur: Done.

xinhaoyuan (approved):

PR #174, opened 2021-07-30
Add coverage support for fuzzing_regression_test

zhenyudg:
Adds support for `bazel coverage` to gather coverage on fuzz targets replayed on the corpus.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -70,3 +70,15 @@ build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=jazzer
 build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_sanitizer=asan
 # Workaround for https://github.com/bazelbuild/bazel/issues/11128
 build:asan-jazzer --//fuzzing:cc_engine_sanitizer=asan
+
+# Coverage with Replay (C/C++ only)
+coverage --//fuzzing:cc_engine=//fuzzing/engines:replay
+coverage --@rules_fuzzing//fuzzing:cc_engine_instrumentation=none
+coverage --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
+coverage --instrument_test_targets
+coverage --action_env=BAZEL_USE_LLVM_NATIVE_COVERAGE=1
+coverage --action_env=GCOV=llvm-profdata-10
+coverage --action_env=BAZEL_LLVM_COV=llvm-cov-10
+coverage --combined_report=lcov
+coverage --experimental_use_llvm_covmap
+coverage --experimental_generate_llvm_lcov
--- a/fuzzing/private/regression.bzl
+++ b/fuzzing/private/regression.bzl
@@ -46,7 +46,14 @@ exec '{engine_launcher}'
     runfiles = ctx.runfiles()
     runfiles = runfiles.merge(ctx.attr.binary[DefaultInfo].default_runfiles)
     runfiles = runfiles.merge(binary_info.engine_info.launcher_runfiles)
-    return [DefaultInfo(executable = script, runfiles = runfiles)]
+
+    return [
+        DefaultInfo(executable = script, runfiles = runfiles),
+        coverage_common.instrumented_files_info(
+            ctx,
+            dependency_attributes = ["binary"],
+        ),
+    ]
 
 fuzzing_regression_test = rule(
     implementation = _fuzzing_regression_test_impl,
@@ -60,6 +67,16 @@ Executes a fuzz test on its seed corpus.
             providers = [FuzzingBinaryInfo],
             cfg = "target",
             mandatory = True,
+        ),
+        "_lcov_merger": attr.label(
+            default = "@bazel_tools//tools/test:lcov_merger",
+            executable = True,
+            cfg = "host",
+        ),
+        "_collect_cc_coverage": attr.label(
+            default = "@bazel_tools//tools/test:collect_cc_coverage",
+            executable = True,
+            cfg = "host",
         ),
     },
     test = True,
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -81,6 +81,28 @@ jobs:
           bazel test --verbose_failures --test_output=all \
               --build_tag_filters=fuzz-test --config=${{ matrix.config }} \
               //examples:all
+  coverage:
+    name: Coverage gathering (C++)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        target: ["//examples:empty_fuzz_test", "//examples:re2_fuzz_test"]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -yq \
+            clang-10 \
+            libunwind-dev \
+            libblocksruntime-dev
+      - name: Gather coverage
+        run: |
+          bazel coverage ${{ matrix.target }}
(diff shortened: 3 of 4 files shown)


fuzzing/private/regression.bzl
```diff
@@ -48,7 +48,7 @@ exec '{engine_launcher}'
     runfiles = runfiles.merge(binary_info.engine_info.launcher_runfiles)
     return [DefaultInfo(executable = script, runfiles = runfiles)]
 
-fuzzing_regression_test = rule(
+fuzzing_regression_test_script = rule(
```
stefanbucur, line 51: I'm not sure I understand what is the reason for this change.

.bazelrc
```diff
@@ -70,3 +70,15 @@ build:asan-jazzer --@rules_fuzzing//fuzzing:cc_engine_instrumentation=jazzer
...
+coverage --//fuzzing:cc_engine=//fuzzing/engines:replay
+coverage --@rules_fuzzing//fuzzing:cc_engine_instrumentation=none
+coverage --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
+coverage --instrument_test_targets
+coverage --action_env=BAZEL_USE_LLVM_NATIVE_COVERAGE=1
+coverage --action_env=GCOV=llvm-profdata-10
+coverage --action_env=BAZEL_LLVM_COV=llvm-cov-10
+coverage --combined_report=lcov
+coverage --experimental_use_llvm_covmap
+coverage --experimental_generate_llvm_lcov
```
stefanbucur, line 84: I think this is great for the rules_fuzzing project itself, but what does it mean for the other projects using the fuzzing rules? Will they have to copy-paste and maintain this snippet in their own `.bazelrc` files? Could we integrate these into the existing instrumentation framework somehow, or this is not worth the effort?

fuzzing/private/regression.bzl
zhenyudg: `sh_test` supports coverage gathering on its C/C++ data dependencies (ref: https://youtu.be/P51Rgcbxhyk?t=599).

.bazelrc
zhenyudg: I anticipate that projects would maintain their own configurations for using `bazel coverage`. 

These flags are not specific to rules_fuzzing, but are instead for gathering coverage on clang-compiled code in general, whether from a `cc_test` or a `cc_fuzz_test`.

zhenyudg:
@googlebot I signed it!

.bazelrc
stefanbucur: > I anticipate that projects would maintain their own configurations for using `bazel coverage`.

This might be suboptimal in the long-term. My concern is that if we need to make changes to this set of flags over time (e.g., as the Bazel coverage support evolves), it will be very difficult to propagate them to client libraries if they need to duplicate them in their `.bazelrc` files.

Ideally, we'd be somewhere where we could ask our users to simply define:

```
coverage --@rules_fuzzing//fuzzing:cc_engine=@rules_fuzzing//fuzzing/engines:replay
coverage --@rules_fuzzing//fuzzing:cc_engine_instrumentation=coverage
coverage --@rules_fuzzing//fuzzing:cc_engine_sanitizer=none
```

and things would just work out of the box.

In any case, this is probably good enough for *our own* coverage support, so we can move forward with it.

fuzzing/private/regression.bzl
stefanbucur: That is true, but if we do this, one target becomes two targets, so we add one more target to all regression tests instantiations. I'm wondering if we can avoid that. Can we take any of the steps outlined at https://docs.bazel.build/versions/3.5.0/skylark/rules.html#code-coverage-instrumentation?

stefanbucur (changes requested):
Apologies for the delay!

fuzzing/private/regression.bzl
```diff
@@ -61,6 +68,16 @@ Executes a fuzz test on its seed corpus.
...
+        "_lcov_merger": attr.label(
+            default = "@bazel_tools//tools/test:lcov_merger",
+            executable = True,
+            cfg = "host",
+        ),
+        "_collect_cc_coverage": attr.label(
+            default = "@bazel_tools//tools/test:collect_cc_coverage",
+            executable = True,
+            cfg = "host",
+        ),
```
stefanbucur, line 80: I think I need to understand this better: Why do we need to add these dependencies? Are these going to be built even in non-coverage mode?

(It would help if you could point me to some documentation discussing this.)

zhenyudg: [collect_cc_coverage](https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/tools/test/collect_cc_coverage.sh) runs the gcov/llvm-cov commands to gather coverage.

[lcov_merger](https://github.com/bazelbuild/bazel/tree/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator) merges different coverage output files into a single output file for consumption.

I found these dependencies from the [implementation of the native cc_test rule](https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/main/java/com/google/devtools/build/lib/bazel/rules/cpp/BazelCcTestRule.java#L47).

They will be built in non-coverage mode. Native rules with coverage support use late-bound attributes to avoid building these dependencies unnecessarily, but that's [not yet an option in Starlark](https://github.com/bazelbuild/bazel/issues/8736).

stefanbucur: Unfortunately I'm reluctant to accept this change as-is. I understand this is a Starlark limitation, but the workaround of including these dependencies on all builds is inadequate IMO.

Let me do some additional research to better understand all of our alternatives. If coverage support is not mature enough yet in Bazel, I would postpone the addition of this feature for now.

fmeum: This should be fixed by https://github.com/bazelbuild/bazel/pull/13983. Until then, I don't know of a better solution.

fmeum: [The underlying issue](https://github.com/bazelbuild/bazel/issues/13978) is now a release blocker for Bazel 5, so it's likely that `lcov_merger` will be provided by default in a way that it would be built only with `bazel coverage`. The same does not apply to `collect_cc_coverage`, but that is merely a shell script and thus causes no compilation overhead.

fmeum: My fix [will not be accepted](https://github.com/bazelbuild/bazel/pull/13983#issuecomment-981542636), so this will very likely not be possible with Bazel 5.

fmeum: @zhenyudg I haven't forgotten about this. If you want to know whether there has been progress in getting the underlying issue resolved in Bazel, you can track https://github.com/bazelbuild/bazel/issues/8736.

fmeum: @zhenyudg While working on the upstream fix, I may have found a workaround for Bazel 4+: Use an `alias` as the default for `_lcov_merger` and in that alias, use a `select` on a [`config_setting`](https://docs.bazel.build/versions/main/configurable-attributes.html) for [`collect_code_coverage`](https://docs.bazel.build/versions/main/command-line-reference.html#flag--collect_code_coverage) to switch between the real lcov_merger target and a dummy target (e.g., an empty executable file).

fmeum:
@zhenyudg Are you still planning to work on this PR? I have a concrete need for this feature now and would thus pick it up soon.

zhenyudg:
Superseded by https://github.com/bazelbuild/rules_fuzzing/pull/212

PR #177, opened 2021-08-16
Restore and verify GCC compatibility

### LINKED ISSUE #176: cc_fuzz_test() breaks non-Clang builds
## Actual Behavior

The [docs for cc_fuzz_target](https://github.com/bazelbuild/rules_fuzzing#configuring-the-bazelrc-file) require you to add the following lines to your `.bazelrc`:

```
# Force the use of Clang for C++ builds.
build --action_env=CC=clang
build --action_env=CXX=clang++
```

This makes it impossible to build the repo with any compiler other than Clang. But I want my repo to be compatible with GCC, MSVC, and other compilers, for testing my code across compilers.

I tried specifying these actions for the `libfuzzer` configurations only:

```
# Force the use of Clang for fuzzer builds only.
build:asan-libfuzzer --action_env=CC=clang
build:asan-libfuzzer --action_env=CXX=clang++
```

However, I discovered that the `cc_fuzz_target()` would then fail to build with other compilers. For example, compiling with GCC I got the following error:

```
bazel-out/k8-fastbuild-ST-6de91758680c/bin/external/rules_fuzzing/fuzzing/replay/_objs/replay_main/replay_main.pic.o:replay_main.cc:function main: error: undefined reference to 'LLVMFuzzerInitialize'
```

This is unfortunate; even if nobody tries to build the fuzzing target directly, a broken target like this will make `bazel test ...` fail.

## Expected Behavior

Ideally, `cc_fuzz_target()` would automatically opt itself out of non-Clang builds using [target_compatible_with](https://docs.bazel.build/versions/main/be/common-definitions.html#common.target_compatible_with).

This would cause the fuzz targets to be automatically skipped by Bazel on non-Clang builds. Users could get the best of both worlds: the repo could continue to be compatible with any compiler, but Bazel would only attempt to build the fuzzing rules if the build is using Clang.

I tried doing this myself by adding `target_compatible_with` to my `cc_fuzz_target()` rule:

```
cc_fuzz_test(
    name = "file_descriptor_parsenew_fuzzer",
    srcs = ["file_descriptor_parsenew_fuzzer.cc"],
    deps = [
        "//:descriptor_upb_proto",
        "//:upb",
    ],
    target_compatible_with = [
        "@bazel_tools//tools/cpp:clang",
    ],
)
```

However this didn't work; Bazel seemed to skip the test no matter what I did. I tried both of these commands, and both of them skipped the fuzz test:

```
CC=clang CXX=clang++ bazel test ...
```

and

```
bazel test --compiler=clang ...
```

I'm probably doing something wrong at the Bazel level. If `cc_fuzz_target()` has this built in I wouldn't need to figure out the right incantation to get this working. :)

fmeum:
Fixes #176.

### DIFF
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -126,3 +126,14 @@ jobs:
           bazel test --action_env=CC=gcc --action_env=CXX=g++ \
               --verbose_failures --test_output=all \
               //examples/...
+  regression_tests_mac:
+    name: Regression tests (macOS)
+    runs-on: macos-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Run regression tests on macOS
+        run: |
+          bazel test --verbose_failures --test_output=all \
+              -- //examples/... -//examples/java:all
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -114,3 +114,15 @@ jobs:
           bazel test --verbose_failures --test_output=all \
               --build_tag_filters=fuzz-test --config=${{ matrix.config }} \
               //examples/java/...
+  regression_tests_gcc:
+    name: Regression tests (GCC)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Run regression tests with GCC
+        run: |
+          bazel test --action_env=CC=gcc --action_env=CXX=g++ \
+              --verbose_failures --test_output=all \
+              //examples/...
--- a/BUILD
+++ b/BUILD
@@ -11,3 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+config_setting(
+    name = "clang",
+    flag_values = {"@bazel_tools//tools/cpp:compiler": "clang"},
+)
--- a/fuzzing/replay/replay_main.cc
+++ b/fuzzing/replay/replay_main.cc
@@ -22,7 +22,7 @@
 
 extern "C" {
 int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size);
-int LLVMFuzzerInitialize(int* argc, char*** argv) __attribute__((weak_import));
+int LLVMFuzzerInitialize(int* argc, char*** argv) __attribute__((weak));
 }
 
 namespace {
--- a/examples/BUILD
+++ b/examples/BUILD
@@ -58,6 +58,10 @@ cc_fuzz_test(
 cc_fuzz_test(
     name = "fuzzed_data_provider_fuzz_test",
     srcs = ["fuzzed_data_provider_fuzz_test.cc"],
+    target_compatible_with = select({
+        "//:clang": [],
+        "//conditions:default": ["@platforms//:incompatible"],
+    }),
 )
 
 cc_fuzz_test(
--- a/fuzzing/engines/honggfuzz_launcher.sh
+++ b/fuzzing/engines/honggfuzz_launcher.sh
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-command_line="$(readlink -f ${HONGGFUZZ_PATH})"
+command_line="$(python3 -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' ${HONGGFUZZ_PATH})"
 command_line+=("--workspace=${FUZZER_OUTPUT_ROOT}")
 
 if [[ -n "${FUZZER_SEED_CORPUS_DIR}" ]]; then
--- a/fuzzing/engines/jazzer_launcher.sh
+++ b/fuzzing/engines/jazzer_launcher.sh
@@ -16,7 +16,7 @@
 # engine. The launch configuration is supplied by the launcher script through
 # environment variables.
 
-command_line=("$(readlink -f ${FUZZER_BINARY})")
+command_line=("$(python3 -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' ${FUZZER_BINARY})")
 
 # libFuzzer flags (compatible with Jazzer).
 
--- a/fuzzing/engines/libfuzzer_launcher.sh
+++ b/fuzzing/engines/libfuzzer_launcher.sh
@@ -16,7 +16,7 @@
 # libFuzzer engine. The launch configuration is supplied by the launcher
 # script through environment variables.
 
-command_line=("$(readlink -f ${FUZZER_BINARY})")
+command_line=("$(python3 -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' ${FUZZER_BINARY})")
 
 # libFuzzer flags.
 
--- a/fuzzing/engines/replay_launcher.sh
+++ b/fuzzing/engines/replay_launcher.sh
@@ -16,7 +16,7 @@ if (( ! FUZZER_IS_REGRESSION )); then
     echo "NOTE: Non-regression mode is not supported by the replay engine."
 fi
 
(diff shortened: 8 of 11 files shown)


fmeum:
Fixes #176.

README.md
```diff
@@ -71,8 +71,7 @@ It is best to create command shorthands for the fuzzing configurations you will
 
 ```
 # Force the use of Clang for C++ builds.
-build --action_env=CC=clang
-build --action_env=CXX=clang++
+build --repo_env=CC=clang
```
stefanbucur, line 74: Is this change necessary?

If the intent is to highlight GCC support in the README.md file, I would prefer to still recommend using clang by default, because it is a simpler setup. We can still mention that GCC is supported for non-fuzzer configurations.

BUILD
```diff
@@ -11,3 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+config_setting(
+    name = "clang",
+    flag_values = {"@bazel_tools//tools/cpp:compiler": "clang"},
```
stefanbucur, line 17: Just to confirm: Is this value being affected by setting the `CC` and `CXX` environment vars?

README.md
```diff
@@ -168,8 +167,7 @@ To use Jazzer, it is convenient to also define a `.bazelrc` configuration, simil
 
 ```
 # Force the use of Clang for all builds (Jazzer requires at least Clang 9).
-build --action_env=CC=clang
-build --action_env=CXX=clang++
+build --repo_env=CC=clang
```
stefanbucur, line 170: Same here.

fuzzing/replay/replay_main.cc
```diff
@@ -22,7 +22,7 @@
 
 extern "C" {
 int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size);
-int LLVMFuzzerInitialize(int* argc, char*** argv) __attribute__((weak_import));
+int LLVMFuzzerInitialize(int* argc, char*** argv) __attribute__((weak));
```
stefanbucur, line 25: It appears that here the code is reverting to the state before #139 - which was problematic because it was breaking MacOS builds. Can you confirm there are no regressions on this front?

fmeum: I think that the relevant change for macOS compatibility was the explicit undefined declaration for `LLVMFuzzerInitialize`, not the change from `weak` to `weak_import` (see https://stackoverflow.com/a/34983229/297261). I don't have access to a Mac at the moment to verify this myself, but added another commit that tests whether the replay engine works on macOS.

fmeum: https://github.com/bazelbuild/rules_fuzzing/pull/177/checks?check_run_id=3354656650 shows that the replay launcher compiles on macOS with the changes from this PR. The regression tests still fail since the launcher scripts rely on `readlink -f`, which is unavailable on macOS, but that is an unrelated issue.

BUILD
fmeum: This flag is set to the value of the `compiler` attribute of the `cc_toolchain`, which will be `"clang"` either if users use a hermetic clang toolchain or rely on the auto-detected host toolchain with `CC` set to `clang` on Linux (see https://github.com/bazelbuild/bazel/blob/f141e149ac3210e417b31aec21afed267cdbb092/tools/cpp/unix_cc_configure.bzl#L480). 

This currently doesn't detect `clang` correctly on macOS if Apple clang is used instead of the LLVM version of clang. If we want the FuzzedDataProvider fuzz test example to run there, we could always enable it on macOS (which will almost always use clang).

README.md
fmeum: It is not necessary to achieve GCC compatibility, it just makes switching compilers more cache-friendly in general. Using `action_env` instead of `repo_env` means that the environment for all actions is changed by this config, rather than just that of repository rules (which are responsible for auto-configuring the C/C++ toolchain). 
Since [`CXX` is not read by Bazel](https://github.com/bazelbuild/bazel/blob/f141e149ac3210e417b31aec21afed267cdbb092/tools/cpp/cc_configure.bzl#L147), the second line appears to be entirely redundant.

If you prefer, I can drop this commit and potentially submit it as a separate PR.

BUILD
stefanbucur: Thanks for clarifying!

README.md
stefanbucur: > It is not necessary to achieve GCC compatibility, it just makes switching compilers more cache-friendly in general. Using `action_env` instead of `repo_env` means that the environment for all actions is changed by this config, rather than just that of repository rules (which are responsible for auto-configuring the C/C++ toolchain).

Interesting, I did not realize these vars are only used during toolchain configuration, and not inside each C++ compilation action.

> Since [`CXX` is not read by Bazel](https://github.com/bazelbuild/bazel/blob/f141e149ac3210e417b31aec21afed267cdbb092/tools/cpp/cc_configure.bzl#L147), the second line appears to be entirely redundant.

Can we assume this will never change? If the standard is to configure both variables, I'd say we should follow that and give the Bazel team the liberty to add CXX support in the future.

> 
> If you prefer, I can drop this commit and potentially submit it as a separate PR.

Yeah, that sounds like a better approach.

README.md
stefanbucur: Resolving in favor of the other thread.

fuzzing/replay/replay_main.cc
stefanbucur: Ugh, and I couldn't find an easy replacement unfortunately.

The best I could come with is replacing `readlink -f "<path>"` with `python -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' "<path>"`. If you get the chance, can you test out this change to see if it makes a difference?

stefanbucur (changes requested):

fuzzing/replay/replay_main.cc
fmeum: That will make the script less compatible with various Linux distributions that by default offer `python3` but not `python`. An alternative could be to ask users to install [`coreutils`](https://formulae.brew.sh/formula/coreutils) and fall back to `greadlink -f` if `readlink -f` fails.

stefanbucur: > That will make the script less compatible with various Linux distributions that by default offer `python3` but not `python`.

Should we use `python3` then, instead? I believe the fuzzing rules anyway have a hard dependency on it through the fuzzer launcher.

fmeum: That's a good point, made the change and it seems to work.

fmeum:
@stefanbucur The CI is broken due to [a regression in Bazel 4.2.0](https://github.com/bazelbuild/bazel/issues/13558#issuecomment-903488624) which came out a few days ago and is now used by both OSS-Fuzz and the Bazel tests. Don't worry, projects *using* rules_fuzzing do not seem to be affected.

fmeum:
There will be a 4.2.1 release that fixes the regression, see https://github.com/bazelbuild/bazel/issues/13890. 

@stefanbucur Could you as the maintainer follow the instructions at https://github.com/bazelbuild/bazel/issues/13890#issuecomment-903857945 to add `rules_fuzzing` to the Bazel release CI?

.github/workflows/bazel_test.yml
```diff
@@ -32,10 +32,10 @@ jobs:
             libblocksruntime-dev
       - name: Run unit tests
         run: |
-          bazel test --test_tag_filters=-fuzz-test --build_tests_only //...
+          bazelisk test --test_tag_filters=-fuzz-test --build_tests_only //...
```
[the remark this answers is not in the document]
fmeum: I removed this commit since Bazel 4.2.1 has meanwhile become the latest stable release.

fmeum:
Now that Bazel 4.2.1, the pipeline passes. 

I had to exclude the Java regression tests from the macOS regression test for now, but will enable them once I have understood where to find the required dylibs. All regression tests are now buildable and runnable with gcc and C++ tests are also compatible with macOS.

stefanbucur (approved):

PR #178, opened 2021-09-27
Check if the non-escaped chars are spaces or printable in "C" locale to match libFuzzer's dictionary parsing.

xinhaoyuan:


stefanbucur (approved):
LGTM modulo the CI-Fuzzing failure in the Java targets.

xinhaoyuan:
CI seems to fail because of https://github.com/google/oss-fuzz/issues/5777. @fmeum would you mind take a look? I think maybe we can fix this locally (e.g. by changing driver names).

fmeum:
> CI seems to fail because of [google/oss-fuzz#5777](https://github.com/google/oss-fuzz/issues/5777). @fmeum would you mind take a look? I think maybe we can fix this locally (e.g. by changing driver names).

Should be fixed by https://github.com/google/oss-fuzz/pull/6538.

xinhaoyuan:
> Should be fixed by [google/oss-fuzz#6538](https://github.com/google/oss-fuzz/pull/6538).

Awesome! Thanks :+1:

PR #186, opened 2021-11-27
Link UBSan C++ runtime with a flag rather than explicitly

fmeum:
This simplifies the oss_fuzz repository rule.


stefanbucur (approved):
This is amazing, was not aware of this option :)

fmeum:
> This is amazing, was not aware of this option :)

Neither was I, only found it while reading LLVM source code.

PR #187, opened 2021-11-28
Support UBSan for local fuzzing

fmeum:
With Jazzer supporting full UBSan as of https://github.com/CodeIntelligenceTesting/jazzer/pull/169 as well as a much simpler way to link the UBSan C++ runtime via the flag used in #186, UBSan can now be supported in local mode without introducing additional complexity.

The list of enabled UBSan checks is taken from OSS-Fuzz.

The commit also adds tests to verify that both C++ and Java fuzz tests support the UBSan C++ checks without linker errors.


stefanbucur (changes requested):
Looks great overall, thanks for contributing! Some naming issues only.

fmeum:
I addressed your changes and expanded some of the comments.

stefanbucur (approved):

PR #192, opened 2022-01-19
Use latest version in WORKSPACE setup blurb.

stefanbucur:


fmeum:
I have an idea what broke here, will look into fixing it.

@stefanbucur Given that the CI Fuzz tests are fundamentally non-hermetic due to them relying on binaries provided by OSS-Fuzz, it could be a good idea to run the CI checks daily (see [this workflow](https://github.com/fmeum/rules_jni/blob/6c457ed979d9e5648b1162b42018509a059cb2e6/.github/workflows/run-tests-externally.yml#L9) for an example). That would make it easier for us to root cause the problems and solve them more timely. What do you think?

stefanbucur:
Replaced by #197

PR #195, opened 2022-01-19
Fix native library handling

fmeum:
See the individual commit messages.


fmeum:
This PR fixes the underlying issue in the Java logic, but fails to provide Bazel 5 support due to a separate issue that only shows up on macOS. I will debug this later.

@stefanbucur The bug is only relevant when fuzzing Java mixed with native libraries. If that's not super urgent to fix, you could hold the 0.3.1 release until I have fixed this Bazel 5 issue. Otherwise, feel free to cherry-pick the first commit.

fmeum:
@stefanbucur This is ready for review now and fixes the two bugs I am aware of.

stefanbucur (approved):

PR #196, opened 2022-01-20
Improve CI setup

fmeum:


fmeum:
@stefanbucur Let me know if you also want me to pin the Bazel version for this repository to Bazel 5. This would affect the version used for non-scheduled, per-PR CI runs, OSS-Fuzz CI Fuzz runs as well as local development.

.github/workflows/bazel_test.yml
```diff
@@ -30,6 +35,9 @@ jobs:
             clang-10 \
             libunwind-dev \
             libblocksruntime-dev
+      - name: Use Bazel last_green on schedule
+        if: github.event.schedule == '0 0 * * *'
```
stefanbucur, line 39: Checking for the specific scheduling pattern is brittle. If the next person changes the schedule, this if statement will silently be dropped.

It is also unreasonable to have to configure the same scheduling pattern in all N places. This should only happen in a single place, at the beginning.

Any ideas of how to fix this?

.github/workflows/bazel_test.yml
```diff
@@ -30,6 +35,9 @@ jobs:
...
+        run: echo "USE_BAZEL_VERSION=last_green" >> $GITHUB_ENV
```
stefanbucur, line 40: I was briefly reading about environment variables for Github actions. Have you considered the alternative of an `env` section at the beginning of the job, guarded by an `if` conditional? Any trade-offs?

stefanbucur (changes requested):

.github/workflows/bazel_test.yml
fmeum: I haven't found any documentation on what `github.event.schedule` will evaluate to if the workflow hasn't been triggered by a schedule. It is also quite difficult to test. Furthermore, GitHub Actions doesn't support YAML anchors: https://github.community/t/support-for-yaml-anchors/16128/4.

Do you have a guess for what the default value of `github.event.schedule` should be?

.github/workflows/bazel_test.yml
fmeum: It's possible to set environment variables via `env`, but I thought that didn't work conditionally: The `if` would run the entire job conditionally, not just set the variables. But I don't know GitHub Actions well enough to say with certainty that it's not possible.

.github/workflows/bazel_test.yml
stefanbucur: > I haven't found any documentation on what `github.event.schedule` will evaluate to if the workflow hasn't been triggered by a schedule. It is also quite difficult to test. Furthermore, GitHub Actions doesn't support YAML anchors: https://github.community/t/support-for-yaml-anchors/16128/4.
> 
> Do you have a guess for what the default value of `github.event.schedule` should be?

`null` perhaps? You can quickly test it by running a `run: exit 1` action on the condition.

stefanbucur: I was looking at https://docs.github.com/en/actions/learn-github-actions/expressions#literals

stefanbucur: Another idea: Unconditionally run `echo ${{ github.event.schedule }}` and see if it prints the value used for non-scheduled runs.

fmeum: Thanks for the idea. I tried `echo "'${{ github.event.schedule }}'"` and got `''`. I updated the PR accordingly, which gets rid of the repeated schedule. 

I haven't found a way to conditionally define an environment variable for an entire workflow though.

.github/workflows/bazel_test.yml
stefanbucur: I found this: https://github.com/actions/runner/issues/409

I'm wondering if this form works at the beginning of the job:

```
jobs:
  unit_tests:
    env:
      USE_BAZEL_VERSION: ${{ github.event.schedule == '' && 'last_green' || '' }}
```

But I'm happy with either form, so let me know what you decide.

stefanbucur (approved):

.github/workflows/bazel_test.yml
fmeum: Great find (and great hack). I verified it works by trying it both this way and with the inverted condition, observing that the Bazel version did indeed change as expected.

Now that we can do this inline in an env, I moved it to the workflow level, which gets rid of all repetitions.

stefanbucur: > Now that we can do this inline in an env, I moved it to the workflow level, which gets rid of all repetitions.

Excellent, thank you!

PR #199, opened 2022-02-01
Update rules_python to 0.6.0

### LINKED ISSUE #198: Use of deprecated `py_repositories`
https://github.com/bazelbuild/rules_fuzzing/blob/fc0e7d437fda6666b9f5f25d1a40ae919cfcecf2/fuzzing/init.bzl#L22

Newer version of `rules_python` have this warning now that is caused by the line above -- also there does not seem to be a way for users of `rules_fuzzing` to turn this off, so I would suggest to remove it.

```
DEBUG: .../external/rules_python/python/repositories.bzl:8:10: py_repositories is a no-op and is deprecated. You can remove this from your WORKSPACE file
```

fmeum:
This allows us to get rid of the py_repositories() call in init.bzl.
This call is no longer needed with newer versions of rules_python and produces warnings.

Fixes #198

### DIFF
--- a/fuzzing/init.bzl
+++ b/fuzzing/init.bzl
@@ -15,11 +15,9 @@
 """Dependency initialization utilities."""
 
 load("@rules_python//python:pip.bzl", "pip_install")
-load("@rules_python//python:repositories.bzl", "py_repositories")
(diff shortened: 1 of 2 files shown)


stefanbucur (approved):
Thank you!

PR #213, opened 2022-10-11
Remove libFuzzer stub source file

fmeum:
The stub containing the warning is redundant: If it is linked in, then the target depending on it will also pick up the `-fsanitize=fuzzer` flag and thus link in libFuzzer.

On macOS, where weak symbols are treated differently by default, the stub can lead to libFuzzer not being linked at all.


stefanbucur (approved):
Thanks!

PR #217, opened 2022-11-21
Use Bazel's default `PATH` when running Python actions

fmeum:
Without `use_default_shell_env`, `PATH` was not set for Python actions, which resulted in `python` (or `python3`) not being found.

### DIFF
--- a/.github/workflows/bazel_test.yml
+++ b/.github/workflows/bazel_test.yml
@@ -32,7 +32,7 @@ env:
 jobs:
   unit_tests:
     name: All unit tests
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


fmeum:
@stefanbucur The fix consists of two parts: the first commit of this PR and https://github.com/google/oss-fuzz/pull/9024, which I mocked with the temporary second commit.

fmeum:
@stefanbucur The upstream change has been reverted, so this is all green now.

stefanbucur (approved):
Thanks! Opened #221 to keep track of the ubuntu-latest issue.

PR #218, opened 2022-11-21
Update and simplify Jazzer integration

fmeum:
This commit updates Jazzer to v0.17.1 and restores compatibility with OSS-Fuzz after its recent Jazzer update.

Since Jazzer can now be run directly as a `java_binary` obtained from Maven, the setup required for `java_fuzz_test` has been simplified:
* Jazzer can now be used without extra WORKSPACE setup.
* The Jazzer runtime is represented as an engine analogous to the case of C++ fuzz engines.
* Since Jazzer now loads the correct sanitizer runtimes at runtime, it's no longer necessary to use different Jazzer binaries depending on the sanitizer setting. Logic is added to collect the runtimes from the C++ toolchain at build time and add them to Jazzer's runfiles.

The Jazzer update also makes rules_fuzzing compatible with Bazel 6.

### DIFF
--- a/fuzzing/engines/BUILD
+++ b/fuzzing/engines/BUILD
@@ -68,6 +68,14 @@ java_fuzzing_engine(
     name = "jazzer",
     display_name = "Jazzer",
     launcher = "jazzer_launcher.sh",
-    library = "@jazzer//agent:jazzer_api_compile_only",
+    library = ":jazzer_stub",
     visibility = ["//visibility:public"],
+)
+
+java_library(
+    name = "jazzer_stub",
+    exports = [
+        "@maven_jazzer//jar",
+        "@maven_jazzer_api//jar",
+    ],
 )
--- a/fuzzing/engines/BUILD
+++ b/fuzzing/engines/BUILD
@@ -72,10 +72,12 @@ java_fuzzing_engine(
     visibility = ["//visibility:public"],
 )
 
+# This wrapper target is needed as Jazzer consists of two separate Java targets,
+# but java_fuzzing_engine's library attribute only accepts a single target.
 java_library(
     name = "jazzer_stub",
     exports = [
-        "@maven_jazzer//jar",
-        "@maven_jazzer_api//jar",
+        "@rules_fuzzing_jazzer//jar",
+        "@rules_fuzzing_jazzer_api//jar",
     ],
 )
--- a/fuzzing/private/runtime/BUILD
+++ b/fuzzing/private/runtime/BUILD
@@ -26,11 +26,13 @@ clang_runtime_lib(
         # LLVM 14 and earlier
         "libclang_rt.asan-x86_64.so",
     ],
+    tags = ["manual"],
 )
 
 clang_runtime_lib(
     name = "asan_macos",
     basenames = ["libclang_rt.asan_osx_dynamic.dylib"],
+    tags = ["manual"],
 )
 
 clang_runtime_lib(
@@ -41,9 +43,11 @@ clang_runtime_lib(
         # LLVM 14 and earlier
         "libclang_rt.ubsan_standalone-x86_64.so",
     ],
+    tags = ["manual"],
 )
 
 clang_runtime_lib(
     name = "ubsan_macos",
     basenames = ["libclang_rt.ubsan_osx_dynamic.dylib"],
+    tags = ["manual"],
 )
--- a/fuzzing/private/runtime/BUILD
+++ b/fuzzing/private/runtime/BUILD
@@ -1,0 +1,49 @@
+load(":clang_runtime_lib.bzl", "clang_runtime_lib")
+
+alias(
+    name = "asan",
+    actual = select({
+        "@platforms//os:linux": ":asan_linux",
+        "@platforms//os:macos": ":asan_macos",
+    }),
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "ubsan",
+    actual = select({
+        "@platforms//os:linux": ":ubsan_linux",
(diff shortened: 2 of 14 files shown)


fmeum:
@stefanbucur Could you review? Let me know if you want me to add more information about what changed in Jazzer and how it affects rules_fuzzing.

fuzzing/repositories.bzl
```diff
@@ -14,19 +14,17 @@
 
 """Contains the external dependencies."""
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_jar")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load("//fuzzing/private/oss_fuzz:repository.bzl", "oss_fuzz_repository")
 
-def rules_fuzzing_dependencies(oss_fuzz = True, honggfuzz = True, jazzer = False):
+def rules_fuzzing_dependencies(oss_fuzz = True, honggfuzz = True, jazzer = True):
```
stefanbucur, line 21: Double-checking: Is it time to flip this to True now?

(The rationale for keeping it False was that the majority of use cases of fuzzing were still for C++ only, and that dual Java and C++ Bazel projects were rather rare. 

In other words, if our users upgraded rules_fuzzing to the current version, in how many cases would the extra new deps would go unused? If that's a majority of users, then I'd argue against flipping this flag.)

fuzzing/private/runtime/clang_runtime_lib.bzl
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Macro that extracts clang runtime libraries from the current cc_toolchain."""
+
+def clang_runtime_lib(*, name, basenames, **kwargs):
+    """Provide the first available clang runtime library with any of the given basenames."""
+    native.genrule(
+        name = name,
+        outs = basenames[:1],
```
stefanbucur, line 21: This part about the output seems less intuitive - perhaps document it in the docstring? 

In particular, it was not clear to me if the output of this rule is a file name or the actual file (library).

fuzzing/private/runtime/BUILD
```diff
@@ -0,0 +1,49 @@
...
+        "libclang_rt.ubsan_standalone.so",
+        # LLVM 14 and earlier
+        "libclang_rt.ubsan_standalone-x86_64.so",
+    ],
+)
+
+clang_runtime_lib(
+    name = "ubsan_macos",
+    basenames = ["libclang_rt.ubsan_osx_dynamic.dylib"],
+)
```
stefanbucur, line 49: Should these rules always be defined? I believe Bazel supports specifying compatibility criteria: https://bazel.build/extending/platforms#skipping-incompatible-targets

fuzzing/engines/BUILD
```diff
@@ -68,6 +68,14 @@ java_fuzzing_engine(
     name = "jazzer",
     display_name = "Jazzer",
     launcher = "jazzer_launcher.sh",
-    library = "@jazzer//agent:jazzer_api_compile_only",
+    library = ":jazzer_stub",
     visibility = ["//visibility:public"],
 )
+
+java_library(
```
stefanbucur, line 75: Maybe explain in a comment the role of this rule? It seems to provide no functionality?

stefanbucur (changes requested):

fuzzing/repositories.bzl
fmeum: This commit makes it so that Java fuzz tests "just work" with `jazzer = True`, which adds two external repos, whereas previously calling Jazzer's own WORKSPACE macros was required additionally. 

I now see the situation as follows: An additional repository with a sufficiently unique name (I pushed a commit to prefix it with `rules_fuzzing`) that is only fetched lazily when the corresponding functionality, in this case `java_fuzz_test`, is used, doesn't really have much of an associated cost. It won't conflict with user-defined deps and won't be evaluated until used.

I of course don't have reliable data on what percentage of rules_fuzzing users uses which language, but based on questions asked on Bazel Slack, I would expect Java to be at least as popular as e.g. honggfuzz, which is also a "cheap" dep in the sense I described above.

fuzzing/private/runtime/clang_runtime_lib.bzl
fmeum: I now explain this in the docs.

fuzzing/engines/BUILD
fmeum: I added a comment.

fuzzing/private/runtime/BUILD
fmeum: It does, but that doesn't factor into target selection, it just improves error messages and makes wildcard builds pass. Here, I had `tags = ["manual"]` in the macro, which was arguably too magic - I moved it out into the BUILD file.

fuzzing/repositories.bzl
stefanbucur: > This commit makes it so that Java fuzz tests "just work" with `jazzer = True`, which adds two external repos, whereas previously calling Jazzer's own WORKSPACE macros was required additionally.

Good point, I had missed this aspect!

> 
> I now see the situation as follows: An additional repository with a sufficiently unique name (I pushed a commit to prefix it with `rules_fuzzing`) that is only fetched lazily when the corresponding functionality, in this case `java_fuzz_test`, is used, doesn't really have much of an associated cost. It won't conflict with user-defined deps and won't be evaluated until used.

Sound good to me!

stefanbucur (approved):

PR #219, opened 2022-11-23
Do not deviate from UBSan default for local fuzzing

fmeum:
OSS-Fuzz uses different compile-time settings for UBSan tuned for their setup, but it doesn't make sense to enforce these also for local users.

In OSS-Fuzz mode, these flags will still be picked up from the environment.


stefanbucur (approved):

stefanbucur:
Waiting for all checks to pass before merging.

fmeum:
@stefanbucur Tests pass after the rebase.

PR #220, opened 2022-11-24
Fix memory leak in Java runfiles tests

fmeum:


examples/runfiles_fuzz_test.cc
```diff
@@ -19,19 +19,20 @@
...
 
 namespace {
-  Runfiles *runfiles = nullptr;
+  std::unique_ptr<Runfiles> runfiles;
 }
 
 extern "C" void LLVMFuzzerInitialize(int *argc, char ***argv) {
   std::string error;
-  runfiles = Runfiles::Create((*argv)[0], &error);
+  runfiles.reset(Runfiles::Create((*argv)[0], &error));
```
stefanbucur, line 35: We're trying to follow the Google C++ style guide, which prohibits non-trivially destructible globals: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

How about declaring the runfiles global static? This should prevent the leak detector from firing.

fmeum: I ran the test again on `master`, leak detection only (rightfully) complains about the Java test. I thus dropped the change to `runfiles_fuzz_test.cc` from this PR.

stefanbucur (approved):

PR #225, opened 2023-03-22
Update rules_python and use new deps installer

keith:


stefanbucur:
Thanks a lot for the contribution! It looks like the new rules_python expects a recent Bazel version, as well. We haven't changed the pinned version in a while; if you'd like to fix that, I'd prefer we do it first in a separate PR. Thanks!

stefanbucur:
> @stefanbucur sounds good, do you want to bump to 5.4 or straight to 6.x?

6.x would be great - but in case we run into unexpected incompatibilities that might be non-trivial to fix, 5.4 would be okay as well.

fmeum:
Updating to 6.0 won't just work because rules_fuzzing has a transitive dependency (through Jazzer) on an older version of rules_kotlin, which isn't compatible with Bazel 6. I have a draft PR that would bump the Jazzer dependency in rules_fuzzing, but that requires (since there is no way to pin it) bumping the Jazzer version used in oss-fuzz, which has to be done very carefully.

@keith If you do need 6.0.0 instead of just 5.4.0, let me know and I will try to come up with a shortcut.

keith:
sounds good, not a huge rush for me i just wanted to do the rules_python bump here

stefanbucur (approved):

keith:
thanks!

PR #229, opened 2023-07-06
Update honggfuzz to 2.5

fmeum:


fmeum:
@stefanbucur I am looking into getting `rules_fuzzing` into the [Bazel Central Registry](https://registry.bazel.build) and that would be easier with a dependency on released versions only, hence the update.

stefanbucur (approved):
Thank you!

PR #233, opened 2023-09-07
Apply buildifier lint fixes

fmeum:


fmeum:
@stefanbucur I don't know why, but it looks like none of the non-lint CI checks are running anymore.

stefanbucur (approved):

stefanbucur:
> @stefanbucur I don't know why, but it looks like none of the non-lint CI checks are running anymore.

Hmm, this is indeed bizarre - I don't see anything changing in our .github directory...

PR #238, opened 2023-11-06
Fix import to include full path

acozzette:
This seems to be necessary for the project to build successfully with Bazel 6.

I also fixed some build errors related to Python 3.12, by upgrading to the most recent versions of absl, six, and rules_python.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -22,6 +22,10 @@ load("@rules_fuzzing//fuzzing:repositories.bzl", "rules_fuzzing_dependencies")
 
 rules_fuzzing_dependencies()
 
+load("@rules_python//python:repositories.bzl", "py_repositories")
+
+py_repositories()
+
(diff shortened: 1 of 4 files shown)


stefanbucur:
Happy to accept this, but I see some regression failures. Do you have a sense what might be going on there?

fmeum:
This could very well be a regression caused by updates to the macOS runner. Probably worth trying to update rules_python to the latest version.

acozzette:
I just pushed a few more fixes that should hopefully address the failure. I think @fmeum is right and this was caused by the MacOS runners recently upgrading to Python 3.12. We hit the same problem in protobuf recently.

I am going to be out for the next four weeks starting on Monday, so if this ends up not being mergeable as is then someone else may want to take over.

stefanbucur (approved):
Thanks a lot for the quick fixes - the tests seem to pass on all platforms now.

PR #240, opened 2023-11-16
Update pip_parse invocation

### LINKED ISSUE #239: rules_fuzzing incompatible with rules_python > 0.26
## Expected Behavior

`rules_fuzzing` is compatible with recent versions of `rules_python`.

## Actual Behavior

`rules_fuzzing_init` calls `pip_parse`, but the signature of that function changed in rules_python 0.13 (specifically  in https://github.com/bazelbuild/rules_python/pull/807). Instead of a single `requirements` attribute, there's now three per-OS attributes (`requirements_linux`, `requirements_windows`, `requirements_darwin`).

This means a project that uses `rules_fuzzing` needs to replace a call to `rules_fuzzing_init` with a reimplementation that looks like,

```
pip_parse(
    name = "fuzzing_py_deps",
    extra_pip_args = ["--require-hashes"],
    python_interpreter_target = interpreter,
    requirements_darwin = "@rules_fuzzing//fuzzing:requirements.txt",
    requirements_linux = "@rules_fuzzing//fuzzing:requirements.txt",
    requirements_windows = "@rules_fuzzing//fuzzing:requirements.txt",
)
```

The fix is simple enough (change the `rules_fuzzing_init` to look like this), but will make `rules_fuzzing` incompatible with versions of `rules_python` older than 0.13. That version is a little over a year old, it came out on September 25, 2022.

I'll send a PR.

## Specifications

- Version: Still broken on the `master` branch.
- Platform: Broken on any platform.

tpudlik:
The `requirements` argument is deprecated and will be removed in the next rules_python release.

Fixes #239

### DIFF
--- a/fuzzing/init.bzl
+++ b/fuzzing/init.bzl
@@ -21,6 +21,6 @@ def rules_fuzzing_init():
     pip_parse(
         name = "fuzzing_py_deps",
         extra_pip_args = ["--require-hashes"],
-        requirements = "@rules_fuzzing//fuzzing:requirements.txt",
+        requirements_lock = "@rules_fuzzing//fuzzing:requirements.txt",
     )
     bazel_skylib_workspace()


stefanbucur (approved):
Thanks!

PR #241, opened 2023-12-12
Update to Bazel 7

fmeum:
See the individual commit messages for details.


fmeum:
@stefanbucur Tests are passing now. This was caused by a mismatch between clang and llvm-cov/llvm-profdata versions.

stefanbucur (approved):

PR #242, opened 2024-01-15
Add support for Bzlmod

### LINKED ISSUE #189: Consider adding support for the bzlmod module system
We should figure out to to expose dependencies like Honggfuzz, and the OSSFuzz repository rules.

See more at https://docs.bazel.build/versions/main/bzlmod.html

fmeum:
Requires fixing a case of a non-hermetic Python import in
`dict_validation_test` that breaks with Python 3.11, with which Bazel uses `PYTHONSAFEPATH` by default.

A follow-up PR will provide the publishing setup for releases to the
Bazel Central Registry (BCR).

Fixes #189


examples/bzlmod/MODULE.bazel
```diff
@@ -0,0 +1,13 @@
+bazel_dep(name = "rules_fuzzing")
+local_path_override(
+    module_name = "rules_fuzzing",
+    path = "../..",
+)
+
+# TODO: Get rid of this.
```
fmeum, line 7: @rickeylev This is what I need to get rid of.

rickeylev: Can you provide context? Is this related to the convo on slack a few days ago?

fmeum:
@stefanbucur This is what's needed to make `rules_fuzzing` seamless to add as a dependency with Bazel's new dependency management system. I can submit a follow-up change that automates releases to the Bazel Central Registry whenever you create a release.

I pushed an additional commit to work around an issue with OSS-Fuzz builds that I think is unrelated to any changes in this PR. Since today (but not yesterday, see https://github.com/bazelbuild/rules_fuzzing/actions/runs/7535376849), there are new OSS-Fuzz failures that I don't know how to work around. Is it possible that the clang@HEAD used by OSS-Fuzz is broken?

MODULE.bazel
```diff
@@ -0,0 +1,57 @@
+module(
+    name = "rules_fuzzing",
+    # Automatically updated during releases by the Publish to BCR app.
+    version = "0.0.0",
+)
+
+bazel_dep(name = "abseil-cpp", version = "20230802.1", repo_name = "com_google_absl")
```
melver, line 7: Abseil has been upgraded since. Does it work with:
```
bazel_dep(name = "abseil-cpp", version = "20240116.0")
```
?

melver (commented):
LGTM, but I don't think I can approve. Is the only blocker here approval?

MODULE.bazel
fmeum: Since Bzlmod uses minimum version selection, rulesets should ideally specify the lowest version they work with to allow users the greatest flexibility. If a higher version of abseil at the same `compatibility_level` breaks rules_fuzzing, that would be considered a bug in the abseil module.

.bazelrc
```diff
@@ -26,7 +22,7 @@ build --cxxopt=-std=c++14
 build --sandbox_tmpfs_path=/tmp
 
 # Strict dependency check for C++ includes.
-build --features=layering_check
```
stefanbucur, line 29: Why no longer make this part of the default build expectations?

stefanbucur: Since I see that `--config=ci` is not used everywhere in bazel_test.yml, let's document here the use of `--config=ci`.

WORKSPACE.bzlmod
```diff
@@ -0,0 +1 @@
+# Intentionally left blank as all deps are defined in MODULE.bazel.
```
stefanbucur, line 1: For my own understanding: Why have this file if we keep it empty? Is it required by the module conventions?

MODULE.bazel
```diff
@@ -0,0 +1,57 @@
+module(
```
stefanbucur, line 1: Add copyright boilerplate.

(Here and everywhere else where there's a new file.)

examples/bzlmod/BUILD.bazel
```diff
@@ -0,0 +1,13 @@
+load("@rules_fuzzing//fuzzing:cc_defs.bzl", "cc_fuzz_test")
```
stefanbucur, line 1: Add copyright boilerplate.

examples/bzlmod/BUILD.bazel
stefanbucur, line 1: Should this be called just `BUILD` for consistency?

stefanbucur (changes requested):
First, my apologies for the latency, I've been largely OOO during the last couple of weeks.

Second, this is great, thanks a lot for doing this migration. I left some comments, largely of cosmetic nature.

I see some CI failures, but otherwise happy to approve this once it's green.

WORKSPACE.bzlmod
fmeum: I added a comment explaining this, it's a rather technical point.

stefanbucur (approved):

stefanbucur:
Approving, but some tests seem to be failing.

fmeum:
@stefanbucur Thanks. I will fix the Bzlmod example shortly, but I'm somewhat clueless about the OSS-Fuzz compilation failure. I'm going to dig deeper later.

fmeum:
I now think that this is a Bazel limitation: The `layering_check` feature relies on sandboxing, which the `build_bazel_fuzz_test` OSS-Fuzz script disables via `--spawn_strategy=standalone`. 

@stefanbucur What do you think of adding `--spawn_strategy=sandboxed` to the OSS-Fuzz test projects for rules_fuzzing as a workaround? This seems better to me than disabling `layering_check`.

fmeum:
I filed https://github.com/bazelbuild/bazel/issues/21592 to track the Bazel issue.

stefanbucur:
> I now think that this is a Bazel limitation: The `layering_check` feature relies on sandboxing, which the `build_bazel_fuzz_test` OSS-Fuzz script disables via `--spawn_strategy=standalone`.

Wow, thanks for digging into this issue - but how did it work before? Is this due to updating the Abseil dependency?

> 
> @stefanbucur What do you think of adding `--spawn_strategy=sandboxed` to the OSS-Fuzz test projects for rules_fuzzing as a workaround? This seems better to me than disabling `layering_check`.

I think this is a good tradeoff indeed - thank you!

fmeum:
The OSS-Fuzz change has been merged and the fuzz tests are now compiling, but they are crashing at startup. I will look into that.

fmeum:
@stefanbucur This is ready now.

stefanbucur (approved):

PR #248, opened 2024-04-12
Fix incompatibility with Bazel 6

fmeum:
Adds a missing `None` to an otherwise failing `getattr`.


stefanbucur (approved):

fmeum:
@stefanbucur Could you create a 0.5.2 release with this fix? Turns out that it does block a few other Google repos from being added to the BCR.

stefanbucur:
Oops, let me do this now.

PR #251, opened 2024-05-15
setup renovate configuration

mmorel-35:
[Renovate](https://docs.renovatebot.com/)  provides a way to receive pr when a new version is available for a dependency.

Only renovate (vs dependabot)  handle bazel dependencies .
This setup configuration dor a weekly check with renovate. It requiees to be activated on the project or organisation level to work correctly .
See [doc](https://docs.renovatebot.com/modules/platform/github/) to finalize

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,5 +3,6 @@
   "extends": [
     "config:recommended",
     "schedule:earlyMondays"
-  ]
+  ],
+  "enabledManagers": ["github-actions"]
 }


stefanbucur:
I think I need more context here - why do we need to support Renovate here?

fmeum:
With Bzlmod and its Minimum Version Selection, it's actually better to *not* update dependencies of "library projects" (such as rulesets). That avoids forcing users to upgrade. 

It would make sense to test against the latest versions of all deps in CI on a regular basis, but that's not covered by the currently available tooling.

stefanbucur:
I'd also err on keeping things as they currently are. We are a small project and dependency management hasn't been an issue so far, but too much dependency churn can increase maintenance cost.

Issue #253, opened 2024-09-12
Replayer ignores hidden corpus files & directories


simonresch:
## Expected Behavior

The replayer should execute all files that are passed as `corpus` to a `cc_fuzz_test` target.

## Actual Behavior

Corpus files in a hidden (`.` prefixed directory) or hidden files are ignored by the replayer as long as the hidden directory/file lies in the bazel project root.

## Steps to Reproduce the Problem

1. Apply the following patch, which adds a corpus file in a hidden directory for the `bzlmod` example:

```
diff --git a/examples/bzlmod/.seed-corpus/foo b/examples/bzlmod/.seed-corpus/foo
new file mode 100644
index 0000000..e69de29
diff --git a/examples/bzlmod/BUILD b/examples/bzlmod/BUILD
index 7333e6c..387c268 100644
--- a/examples/bzlmod/BUILD
+++ b/examples/bzlmod/BUILD
@@ -18,6 +18,7 @@ load("@my_rules_fuzzing//fuzzing:java_defs.bzl", "java_fuzz_test")
 cc_fuzz_test(
     name = "cc_fuzz_test",
     srcs = ["cc_fuzz_test.cc"],
+    corpus = [".seed-corpus/foo"],
 )

 java_fuzz_test(
diff --git a/examples/bzlmod/cc_fuzz_test.cc b/examples/bzlmod/cc_fuzz_test.cc
index b08b002..e6a07b9 100644
--- a/examples/bzlmod/cc_fuzz_test.cc
+++ b/examples/bzlmod/cc_fuzz_test.cc
@@ -16,7 +16,9 @@

 #include <cstddef>
 #include <cstdint>
+#include <iostream>

 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  throw std::runtime_error("Executing corpus file");
   return 0;
 }
```
2. In the `examples/bzlmod` directory run `bazel coverage //:cc_fuzz_test`. 

The replayer should replay the `.seed-corpus/foo` and trigger the inserted exception in `LLVMFuzzerTestOneInput`. Instead, the coverage calculation passes without executing a corpus file. 

## Specifications

- Version: v0.5.2
- Platform: linux amd64

PR #254, opened 2024-09-12
fix(replayer): replay corpus files in hidden dirs

simonresch:
The replayer would not execute hidden corpus files or files in hidden directories if they were in the bazel project root (#253).

### DIFF
--- a/.github/workflows/oss_fuzz.yml
+++ b/.github/workflows/oss_fuzz.yml
@@ -40,7 +40,7 @@ jobs:
         fuzz-seconds: 60
         dry-run: false
     - name: Upload crashes
-      uses: actions/upload-artifact@v1
+      uses: actions/upload-artifact@v4
       if: failure() && steps.build.outcome == 'success'
       with:
         name: artifacts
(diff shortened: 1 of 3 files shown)


fmeum (approved):
The fix looks good to me. Hidden files can always be filtered out with a `glob` in Bazel, which is far more flexible.

Wouldn't hidden files in subdirectories also have been filtered out though? This doesn't look specific to the root (`d_name` should just contain the basename of an entry).

@stefanbucur Could you review this for approval?

simonresch:
> The fix looks good to me. Hidden files can always be filtered out with a `glob` in Bazel, which is far more flexible.
> 
> Wouldn't hidden files in subdirectories also have been filtered out though? This doesn't look specific to the root (`d_name` should just contain the basename of an entry).
> 
> @stefanbucur Could you review this for approval?

Thanks for taking a look @fmeum!

I think hidden files in sub directories works for me since the replayer is started with a prepared corpus directory with file names based on the full path inside the bazel project. E.g. `src/.hidden/foo` end up in the corpus directory as `src-.hidden-foo` which was not filtered out (see https://github.com/bazelbuild/rules_fuzzing/blob/691c8938ff87681f7281cc4ba840e74e3b63c54b/fuzzing/tools/make_corpus_dir.py#L63).

stefanbucur (approved):
Thanks for the fix!

simonresch:
@stefanbucur I added a second commit to update the `upload-artifact` action. This should get the OSS Fuzz github workflow going again. Could you take another look?

PR #261, opened 2025-02-03
Mark extension as reproducible

cerisier:
This removes an unnecessary entry from the lockfiles of rules_fuzzing users.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -18,6 +18,7 @@ module(
     version = "0.0.0",
 )
 
+bazel_dep(name = "bazel_features", version = "1.20.0")
 bazel_dep(name = "abseil-cpp", version = "20240116.1", repo_name = "com_google_absl")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "0.0.8")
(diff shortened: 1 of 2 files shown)


stefanbucur (approved):

stefanbucur:
@cerisier would you like to rebase to fix the build failures?

cerisier:
Sure! I'll do that quickly !

cerisier:
@stefanbucur done

fmeum:
@stefanbucur Could you create a release after this has been merged?

stefanbucur:
> @stefanbucur Could you create a release after this has been merged?

Sure!

fmeum:
@stefanbucur The release to the BCR failed because the release archive was created after the release was published. I can send a manual release.

stefanbucur:
> @stefanbucur The release to the BCR failed because the release archive was created after the release was published. I can send a manual release.

Oh this is interesting! I had only created the release in a single step and I presume the archive was auto-generated along the way. Should I have done it differently?

fmeum:
The auto-generated archives are not a good replacement for release archives as they may change over time, even if the release isn't modified. 

Instead, for a BCR release to work, you need to keep your release in draft state, then add an archive matching the format specified in https://github.com/bazel-contrib/rules_fuzzing/blob/master/.bcr/source.template.json, then publish the release. This could be automated with a workflow: https://github.com/bazel-contrib/rules-template/blob/main/.github/workflows/release.yaml

PR #263, opened 2025-03-28
Support for using honggfuzz with gcc

simonresch:
Adds the flag `--@rules_fuzzing//fuzzing:compiler_type` with possible configurations `cmake` and `gcc`. This allows running honggfuzz with a gcc toolchain which can be useful for code bases that don't compile with clang or where no clang toolchain is available.

Example usage:

```
bazel run //examples:re2_fuzz_test --config asan-honggfuzz-gcc
```

Invalid engine / compiler combination:

```
$ bazel run //examples:re2_fuzz_test --config asan-libfuzzer --@rules_fuzzing//fuzzing:compiler_type=gcc
WARNING: Build options --//fuzzing:cc_engine and --//fuzzing:cc_engine_instrumentation have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
ERROR: Traceback (most recent call last):
	File "/home/simon/git/rules_fuzzing/fuzzing/private/binary.bzl", line 67, column 13, in _fuzzing_binary_transition_impl
		fail("unsupported engine instrumentation '%s' for compiler '%s'" % (instrum_config, compiler_type))
Error in fail: unsupported engine instrumentation 'libfuzzer' for compiler 'gcc'
```

### DIFF
--- a/fuzzing/BUILD
+++ b/fuzzing/BUILD
@@ -73,6 +73,22 @@ bool_flag(
     visibility = ["//visibility:public"],
 )
 
+string_flag(
+    name = "compiler_type",
+    build_setting_default = "clang",
+    values = [
+        "clang",
+        "gcc",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "is_gcc",
+    flag_values = {":compiler_type": "gcc"},
+    visibility = ["//visibility:public"],
+)
+
 exports_files([
     "cc_defs.bzl",
     "java_defs.bzl",
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Force the use of Clang for all builds.
+# Set clang as the default compiler for most configurations.
 build --repo_env=CC=clang
 
 # Needed for abseil-cpp until https://github.com/bazelbuild/bazel/pull/19794 is released.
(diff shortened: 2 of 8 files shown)


simonresch:
Ping @stefanbucur. Is there any interest in adding this?

fuzzing/BUILD
```diff
@@ -73,6 +73,22 @@ bool_flag(
...
+    build_setting_default = "clang",
+    values = [
+        "clang",
+        "gcc",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "is_gcc",
```
fmeum, line 87: This could be replaced with the predefined settings in `@rules_cc//cc/compiler`. Then we don't need `compiler_type` at all as the choice will be determined by the C++ toolchain.

simonresch: I originally attempted something based on `@rules_cc//cc/compiler` but got stuck on this error

```
ERROR: .../rules_fuzzing/examples/java/BUILD:31:15: On dependency edge //examples/java:EmptyFuzzTest_bin (3b35bd8) -|binary|-> //examples/java:EmptyFuzzTest_raw_: attempting to transition on '@@rules_cc+//cc/compiler:gcc' which is not a build setting
```

when adding `@rules_cc//cc/compiler:gcc` as an input to the `fuzzing_binary_transition` transition. Do you have an idea how to access the compiler type from the transition without it being an explicit flag?

fmeum: I think I have something that works, but I'm running into unrelated issues with the rules_cc update (stardoc targets are painful). I'll take another look tomorrow.

fmeum: PTAL at https://github.com/bazel-contrib/rules_fuzzing/pull/288

simonresch:
Closing in favor of https://github.com/bazel-contrib/rules_fuzzing/pull/288

PR #286, opened 2026-01-27
Update deps and bump Python versions

fmeum:
Also mark honggfuzz as Linux-only.


simonresch (approved):
👍 

Sorry, this slipped my notifications.

smcallis:
How often do you all cut releases so we can use this one?

fmeum:
Just started https://github.com/bazel-contrib/rules_fuzzing/actions/runs/24536144819

PR #288, opened 2026-03-18
Support for using honggfuzz with gcc

fmeum:
This allows running honggfuzz with a gcc toolchain which can be useful for code bases that don't compile with clang or where no clang toolchain is available.

Example usage:

```
bazel run //examples:re2_fuzz_test --config asan-honggfuzz
```


fmeum:
I haven't tested the honggfuzz setup

simonresch (approved):
Clever! Works great.

Thanks for helping out with this. :)

PR #289, opened 2026-04-06
prevent symlink loop recursion in directory traversal

jmestwa-coder:
## Summary
Prevent unbounded recursion in directory traversal caused by symlink loops.

## Problem
Traversal currently follows symlinked directories (via `stat`), which can create cycles (e.g., `dirA → dirB → dirA`) and lead to unbounded recursion and unstable behavior.

## Fix
- Use `lstat` to inspect paths without following symlinks  
- Recurse only into real directories (`S_ISDIR && !S_ISLNK`)  
- Avoid entering symlinked directory cycles  

## Behavior
- Regular files and directories are unaffected  
- Symlink entries are still reported but not traversed  
- Traversal is now bounded and deterministic  

## Test
- Added a regression test with a symlink loop (`dirA ↔ dirB`)  
- Verifies traversal runs and remains bounded (`> 0` and `<= 6`)  
- Skips safely if symlinks are unsupported  

## Scope
- Limited to traversal logic in `file_util.cc`  
- No changes to APIs, file handling, or build configuration

### DIFF
--- a/fuzzing/replay/file_util.cc
+++ b/fuzzing/replay/file_util.cc
@@ -21,7 +21,9 @@
 
 #include <cerrno>
 #include <cstdio>
+#include <set>
 #include <string>
+#include <utility>
 
 #include "absl/functional/function_ref.h"
 #include "absl/status/status.h"
@@ -36,7 +38,36 @@ namespace {
 
 absl::Status TraverseDirectory(
     absl::string_view path,
-    absl::FunctionRef<void(absl::string_view, const struct stat&)> callback) {
+    absl::FunctionRef<void(absl::string_view, const struct stat&)> callback,
+    std::set<std::pair<dev_t, ino_t>>& visited);
+
+absl::Status YieldFilesInternal(
+    absl::string_view path,
+    absl::FunctionRef<void(absl::string_view, const struct stat&)> callback,
+    std::set<std::pair<dev_t, ino_t>>& visited) {
+  struct stat path_stat;
+  if (stat(std::string(path).c_str(), &path_stat) < 0) {
+    return ErrnoStatus(absl::StrCat("could not stat ", path), errno);
+  }
+  callback(path, path_stat);
+  if (S_ISDIR(path_stat.st_mode)) {
+    auto dir_id = std::make_pair(path_stat.st_dev, path_stat.st_ino);
+    if (!visited.count(dir_id)) {
+      // Prevent infinite recursion by tracking visited directories (dev,inode).
+      visited.insert(dir_id);
+      absl::Status status = TraverseDirectory(path, callback, visited);
+      if (!status.ok()) {
+        return status;
+      }
+    }
+  }
+  return absl::OkStatus();
+}
+
+absl::Status TraverseDirectory(
+    absl::string_view path,
+    absl::FunctionRef<void(absl::string_view, const struct stat&)> callback,
+    std::set<std::pair<dev_t, ino_t>>& visited) {
   DIR* dir = opendir(std::string(path).c_str());
   if (!dir) {
     return ErrnoStatus(absl::StrCat("could not open directory ", path), errno);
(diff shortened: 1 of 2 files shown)


fuzzing/replay/file_util.cc
```diff
@@ -70,10 +70,12 @@ absl::Status YieldFiles(
...
     absl::FunctionRef<void(absl::string_view, const struct stat&)> callback) {
   struct stat path_stat;
-  if (stat(std::string(path).c_str(), &path_stat) < 0) {
+  // Use lstat so symlink metadata is visible for recursion decisions.
+  if (lstat(std::string(path).c_str(), &path_stat) < 0) {
     return ErrnoStatus(absl::StrCat("could not stat ", path), errno);
   }
-  if (S_ISDIR(path_stat.st_mode)) {
+  // Do not recurse into symlinked directories to avoid traversal loops.
+  if (S_ISDIR(path_stat.st_mode) && !S_ISLNK(path_stat.st_mode)) {
```
simonresch: Thanks for the contribution.

This fixes walking symlink cycles, but it also changes the metadata passed to the callback from `stat()` target metadata to `lstat()` symlink metadata. The [replayer filters for regular files](https://github.com/bazel-contrib/rules_fuzzing/blob/main/fuzzing/replay/test_replayer.cc#L53) which would now skip symlinked files which is likely not intended.

I'm also not sure we want to drop support for symlinked directories entirely. A more compatible fix would likely keep track of visited directories and use that to break symlink cycles without changing normal symlink handling.

simonresch (changes requested):

fuzzing/replay/file_util.cc
jmestwa-coder: Thanks for the feedback 

I’ve updated the patch to keep using stat() so metadata semantics are preserved, and symlinked files/directories continue to behave as before. Instead of skipping symlinks, the fix now tracks visited directories using (st_dev, st_ino) to prevent cycles while still allowing traversal.

jmestwa-coder:
Hi @simonresch  any updates on this PR? Thanks!

fuzzing/replay/file_util.cc
```diff
@@ -36,7 +38,36 @@ namespace {
...
+
+absl::Status YieldFilesInternal(
+    absl::string_view path,
+    absl::FunctionRef<void(absl::string_view, const struct stat&)> callback,
+    std::set<std::pair<dev_t, ino_t>>& visited) {
+  struct stat path_stat;
+  if (stat(std::string(path).c_str(), &path_stat) < 0) {
+    return ErrnoStatus(absl::StrCat("could not stat ", path), errno);
+  }
+  callback(path, path_stat);
```
simonresch: `callback` is here called unconditionally including on directories. This likely breaks existing behavior and fails the test for the file utility functions. You can run them on your end with the command `bazel test //fuzzing/replay:file_util_test` or  `bazel test //...` to run all tests.

The direction to detect cycles on a pair of `path_stat.st_dev, path_stat.st_ino` makes sense to me.

fuzzing/replay/file_util_test.cc
```diff
@@ -113,6 +116,31 @@ TEST(YieldFilesTest, YieldsHiddenFilesAndDirs) {
...
+  }
+  const std::string link_to_a = absl::StrCat(dir_b, "/toA");
+  ASSERT_EQ(symlink(dir_a.c_str(), link_to_a.c_str()), 0);
+
+  std::vector<std::string> collected_paths;
+  const absl::Status status =
+      YieldFiles(root_dir, CollectPathsCallback(&collected_paths));
+  EXPECT_TRUE(status.ok());
+  EXPECT_GT(collected_paths.size(), 0);
+  EXPECT_LE(collected_paths.size(), 6);
```
simonresch: The tests would need one or two normal files after adjusting the callback handling. Also, is there a reason to assert a range of size values instead of the exact number of expected files?

simonresch (changes requested):

jmestwa-coder:
@simonresch, Thanks for the review.
I've updated the patch to preserve the original callback behavior while using `(st_dev, st_ino)` tracking to break cycles without changing symlink handling. I also updated the regression test to use regular files and exact expected results.

PR #290, opened 2026-04-08
Fix path flattening collisions and make runfiles resolution deterministic in make_corpus_dir

jmestwa-coder:
## Summary
Fixes path flattening collisions and makes runfiles resolution deterministic in `make_corpus_dir`.

---

## Problem

- Different inputs could map to the same output (e.g., `dir/../a.txt` vs `a.txt`, `./a.txt` vs `dot-a.txt`)
- Runfiles resolution could pick the wrong script depending on manifest order

---

## Solution

- Preserve input path structure during flattening (no unintended normalization)
- Ensure unique output names using deterministic suffixing and fallback aliasing
- Make runfiles resolution deterministic:
  - prefer workspace entries
  - use `_main` only as fallback
  - ignore unrelated repos

---

## Result

- Distinct inputs always produce distinct outputs
- No silent collisions or overwrites
- Deterministic behavior across runfiles and manifest modes

---

## Tests

- Added coverage for path edge cases and collision scenarios
- Verified runfiles resolution across:
  - `TEST_SRCDIR`
  - `RUNFILES_DIR`
  - `RUNFILES_MANIFEST_FILE`
  - multi-repo manifests

### DIFF
--- a/fuzzing/tools/BUILD
+++ b/fuzzing/tools/BUILD
@@ -83,3 +83,10 @@ py_test(
     srcs = ["dict_validation_test.py"],
     deps = [":dict_validation"],
 )
+
+py_test(
+    name = "make_corpus_dir_test",
+    srcs = ["make_corpus_dir_test.py"],
+    data = ["make_corpus_dir.py"],
+    deps = [requirement("absl-py")],
+)
--- a/fuzzing/tools/make_corpus_dir.py
+++ b/fuzzing/tools/make_corpus_dir.py
@@ -36,12 +36,36 @@ flags.DEFINE_string("output_dir", None, "The path of the output directory")
 
 flags.mark_flag_as_required("output_dir")
 
+def flatten_corpus_path(corpus):
+    prefix = ""
+    if corpus.startswith("./") or (os.sep == "\\" and corpus.startswith(".\\")):
+        prefix = "dot-"
+        corpus = corpus[2:]
+
+    if os.sep == "\\":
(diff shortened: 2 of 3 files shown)


jmestwa-coder:
Hi @fmeum, @simonresch  Can you kindly review this PR. Thank you!

simonresch:
The only caller of `make_corpus_dir` is in `fuzzing_corpus` where https://github.com/bazel-contrib/rules_fuzzing/blob/main/fuzzing/private/common.bzl#L86 calls it with a list of workspace relative, cleaned list of files. Where is the potential for `dir/../a.txt` vs `a.txt`, `./a.txt` ambiguity?

Could you make sure the added test passes? You can run it with the command `bazel test //fuzzing/tools:make_corpus_dir_test`.

jmestwa-coder:
@simonresch.
 I traced the caller path through `fuzzing_corpus` and confirmed that the supported rule path passes normalized workspace-relative `File.path` values from `ctx.files.srcs`, so the `./`, `../`, and absolute-path scenarios are not reachable there.

I've narrowed the change and updated the tests to focus on a reachable collision in the current flattening logic. For example:

* `corpus/a/b.txt`
* `corpus/a-b.txt`

Both currently flatten to `corpus-a-b.txt`, causing `make_corpus_dir` to fail with `ERROR: file ... existed.` and return a non-zero exit code.

The updated tests exercise the real `--corpus_list_file` caller path and cover reachable flattening-collision cases only.

